PR #7, opened 2020-08-16
Support for evaluation using current context.

monkeez:
Support for the Maya 2018 and above feature of [evaluating the current context](https://help.autodesk.com/cloudhelp/2018/ENU/Maya-SDK/files-to-wrap/GUID-A88F1AEC-3B1F-45DC-91C1-61CCF5A7F51A.htm).

That help page explains it well, but it allows you to set contexts and then every time you're calling a command that accepts a context it will automatically evaluate the current context:

```python
context = om.MDGContext(om.MTime(50, unit=om.MTime.uiUnit()))
context.makeCurrent()
node["worldMatrix"][0].asMatrix() # evaluates the context at frame 50
om.MDGContext.kNormal.makeCurrent()
```
Of course that's a silly example where you could just call `node["worldMatrix"][0].asMatrix(50)`. The part where it gets nice is if you wrap that in a context manager and you're querying a bunch of nodes, it helps keep your code clean and avoids the creation of the same context over and over:

```python
with Context(50):
    node1["worldMatrix"][0].asMatrix()
    node2["worldMatrix"][0].asMatrix()
    node3["worldMatrix"][0].asMatrix()
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -412,7 +412,7 @@ mobj = fn.create("transform")
 node = cmdx.Node(mobj)
 ```
 
-This will use the hash if a `cmdx` instance of this `MObject` already exist, else it will instantiate a new. The performance difference is slim and as such this is the recommended approach. The exception is if you happen to already has either an `MObjectHandle` or a corresponding `hashCode` at hand, in which case you can save a handful of cycles per call by using `fromHash` or `fromHex`. 
+This will use the hash if a `cmdx` instance of this `MObject` already exist, else it will instantiate a new. The performance difference is slim and as such this is the recommended approach. The exception is if you happen to already has either an `MObjectHandle` or a corresponding `hashCode` at hand, in which case you can save a handful of cycles per call by using `fromHash` or `fromHex`.
 
 <br>
 
@@ -481,7 +481,7 @@ TypeError: listConnections() got an unexpected keyword argument 'source'
 ```
 
 
-The reason for this limitation is because the functions `cmds` 
+The reason for this limitation is because the functions `cmds`
 
 - Submit an [issue](issues) or [pull-request](#fork) with commands you miss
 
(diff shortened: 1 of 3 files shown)


mottosso:
Oh snap! How I have been looking for this! My understanding was this only worked from C++, with the newer [MDGContextGuard](http://help.autodesk.com/view/MAYAUL/2018/ENU/?guid=__cpp_ref_class_m_d_g_context_guard_html). This is amazing, thanks for this!

Would you be so kind as to also implement a test for this? Either as part of the docstring (they are all executed as actual tests that must pass) or as a [standalone test](https://github.com/mottosso/cmdx/blob/master/tests.py)? Docstring would be best, as it is then also included in the [API docs](https://mottosso.com/cmdx/).

It would also be much preferred to have this handled by Maya 2017 and below, as a no-op would be fine. So long as it doesn't cause errors on 2017 that wouldn't appear on 2018.

monkeez:
Would I put the docstring test in `Plug.asMatrix()`? Because I also changed `_plug_to_python()` which is used by `Plug.read()`, neither of which have docstrings currently.

Sorry, I don't quite understand what you mean about making it be a no-op for Maya 2017 and below. Would you be able to explain?

Thank you!

monkeez:
Also, thank you for this module. It's really cool and something I wish existed for a long time.

mottosso:
Oo, yes good question. It looks like `Plug.read` is sorely missing some examples. How about there?

> Sorry, I don't quite understand what you mean about making it be a no-op for Maya 2017 and below

Yes, got a little carried away. xD We'll need to handle the case of Maya not supporting the feature. Which, looking at it again, should amount to `makeCurrent` not existing on MDGContext in Maya <2018? That should suffice as a signal to users interested in it, so don't worry about that.

For a doctest, since it only exists in 2018+, you'll need to include a e.g. `if hasattr(om, "makeCurrent"):` or the like somewhere to make that not fail on Maya <2018, and to signal to readers of the docs what's required.

What would also be great (I know your TODO list just keeps on growing xD) would be to include your example up above in the README, under the [**Time**](https://github.com/mottosso/cmdx#time) chapter.

I've tinkered with the test suite today to include Maya 2019 and 2020, once we've got a test for this PR we'll be able to spot how it handles from 2015 up till 2020.

> Also, thank you for this module. It's really cool and something I wish existed for a long time.

My pleasure, I also wished it. All the way into existence. xD

monkeez:
Cool thanks, I'll get on these.

Do you think an MDGContext context manager would be useful (like in my example), or is it outside the scope of this project? If you're interested I'll submit a separate pull request for it.

mottosso:
I think it's definitely relevant and useful, happy to see a PR for it.

monkeez:
Sorry for the delays. I couldn't find a way to cleanly add a time based test that works in all Maya versions. I always had to capture some `cmds` returns which made it look messy.

mottosso:
No problem, I think this looks good. A bit unfortunate about the line endings changing on some seemingly unmodified lines, but I don't think it should cause any further issues.

Once you're happy, I'll merge this. Next step would be that context manager, that seems like a really neat usability feature.

monkeez:
Thanks, I'm happy for you to merge this.

Yeah I'm not sure about those line endings, must have been some auto formatting in VSCode. Speaking of which, have you considered using [black](https://github.com/psf/black) to format your python code? I understand that not everyone is fond of it but I've found it useful on my own projects.

I think I'll make an issue for the context manager as I had some questions about the API, there's a few ways I could do it.

mottosso:
> Speaking of which, have you considered using black to format your python code?

It wouldn't format Markdown files though.

> Thanks, I'm happy for you to merge this.

Let's do it!

mottosso:
Thanks @monkeez! I think I'll save on making a new release until we've got the `cmdx.Context` counterpart to really show it off in documentation.

PR #14, opened 2020-09-24
Added DGContext class

monkeez:
First pass of the DGContext context manager. Let me know if you have any suggestions.

### DIFF
--- a/cmdx.py
+++ b/cmdx.py
@@ -16,7 +16,7 @@ from maya import cmds
 from maya.api import OpenMaya as om, OpenMayaAnim as oma, OpenMayaUI as omui
 from maya import OpenMaya as om1, OpenMayaMPx as ompx1, OpenMayaUI as omui1
 
-__version__ = "0.4.4"
+__version__ = "0.4.5"
 
 PY3 = sys.version_info[0] == 3
 
--- a/cmdx.py
+++ b/cmdx.py
@@ -2326,7 +2326,7 @@ class Plug(object):
         """
 
         if time is not None:
-            context = om.MDGContext(om.MTime(time, om.MTime.uiUnit()))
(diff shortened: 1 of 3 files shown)


cmdx.py
```diff
@@ -3917,6 +3914,37 @@ def parent(self, node, parent=None):
...
+
+        if time is not None:
+            if isinstance(time, (int, float)):
+                time = om.MTime(time, om.MTime.uiUnit())
+            super(DGContext, self).__init__(time)
+        else:
+            super(DGContext, self).__init__()
+        self._previousContext = None
+
+    if __maya_version__ >= 2018:
```
mottosso, line 3938: Hm, will this not break any call to..

```python
with cmdx.DGContext(50):
   # ...
```

If the same code is run on Maya 2017?

An alternative might be to keep the `__enter__` and `__exit__` but to call `cmds.warning("Sorry dude, this no work")` or something like that. Or is there `cmds.error` too? Something loud! :D

mottosso:
I like it.

My initial reaction was that `cmds.Context` would have been cleaner.. and maybe it's worth adding a `Context = DGContext` as another alias below, since there is no other jargon for context than a DG context so seems a bit wasteful. But it would be consistent if you're already familiar with it, like `cmdx.Vector` and `cmdx.TransformationMatrix` etc. So we shoul definitely keep it in.

cmdx.py
monkeez: Yeah that's correct. I've just pushed an update, is that what you had in mind?

mottosso: Yes, that's it. Thanks

mottosso:
I think this looks good, merging as soon as the (needless) tests passes (for the README badges to not break).

PR #18, opened 2020-09-26
Added time argument to a number of functions

### LINKED ISSUE #15: getAttr, Plug.asDouble, and Plug.asVector don't support evaluation in another context
I noticed these don't take the time argument, I wasn't sure if there was a particular reason for this?

monkeez:
Resolves #15 

Also fixes asMatrix to use cmdx.DGContext instead of om.MDGContext, otherwise the call will expect an MTime instead of a float.

### DIFF
--- a/cmdx.py
+++ b/cmdx.py
@@ -2289,7 +2289,7 @@ class Plug(object):
     def count(self):
         return self._mplug.evaluateNumElements()
 
-    def asDouble(self):
+    def asDouble(self, time=None):
         """Return plug as double (Python float)
 
         Example:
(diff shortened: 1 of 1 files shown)


monkeez:
As an aside, I was wondering what the unused "type" arguments in cmdx.getAttr and cmdx.setAttr are for?

mottosso:
Nice one.

The unused type is for backwards compatibility with `cmds.getAttr`. It's what lets you swap out the `s` for `x` in most cases and everything would "just work (tm)".

Ready to merge?

monkeez:
Yup, ready to merge!

PR #22, opened 2020-10-01
Automatically convert from string -> cmdx for convenience functions

fvbehr:
Added 
`if isinstance(node, str):`
    `node = encode(node)`
to listRelatives so it behaves more like the cmds command.

### DIFF
--- a/cmdx.py
+++ b/cmdx.py
@@ -993,7 +993,7 @@ class Node(object):
 
         if isinstance(attr, str):
             node, attr = attr.rsplit(".", 1)
-            node = cmdx.encode(node)
+            node = encode(node)
             attr = node[attr]
 
         if isinstance(attr, _AbstractAttribute):
@@ -4176,7 +4176,7 @@ def setAttr(attr, value, type=None):
 
     if isinstance(attr, str):
         node, attr = attr.rsplit(".", 1)
-        node = cmdx.encode(node)
+        node = encode(node)
         attr = node[attr]
 
     attr.write(value)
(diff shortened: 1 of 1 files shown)


mottosso:
Thanks @fvbehr, I think this looks good.

For consistency with the remaining convenience functions..

```py
>>> cmdx.addAttr("pCube1", longName="myAttr", defaultValue=5.0, attributeType="double")
>>> cmdx.connectAttr("pCube1.myAttr", benji + ".tx")
>>> cmdx.setAttr("pCube1.myAttr", 5)
>>> cmdx.delete("pCube1")
```

Do you think we should make the same automatic conversion there? For attributes, we could probably get away with something like..

```py
if isinstance(attr, str):
  node, attr = attr.rsplit(".", 1)
  node = cmdx.encode(node)
  attr = node[attr]
```

mottosso:
For most functions, especially those meant for use by others, there's a docstring with an "Examples" section.

- Like [this one](https://github.com/mottosso/cmdx/blob/d2203e8eee51857b6e1bb1ed175d0ff928854c47/cmdx.py#L4245).

Those examples are actually run and validated as tests. Could you add a few lines to take the new feature for a spin?

fvbehr:
> Do you think we should make the same automatic conversion there?

Yes. That would make it more consistent.

mottosso:
Agreed, will you update the PR? Better to not introduce inconsistency to begin with.

fvbehr:
> Those examples are actually run and validated as tests. Could you add a few lines to take the new feature for a spin?

The current example does not return True at the moment. `cmds.listRelatives` does.
`cmds.createNode` returns unicode and it seems that `cmdx.listRelatives` is expecting a string.
If I add `child = str(child)`
`cmdx.listRelatives(child, parent=True)` returns **[|transform1]**.
Without it, it returns None.

The comparison seems to fail, because `[parent]` is **[u'transform1']** and not **[|transform1]**.

mottosso:
> The current example does not return True at the moment. 

Hm, it does on my machine.

```py
parent = cmdx.createNode("transform")
child = cmdx.createNode("transform", parent=parent)
assert cmdx.listRelatives(child, parent=True) == [parent]
```

Can you show me how you got anything other than `True`?

> The comparison seems to fail, because [parent] is [u'transform1'] and not [|transform1].

Could you post a reproducible?

fvbehr:
> Hm, it does on my machine.

Ah, I see. I was using `cmds.createNode`. That fails. When using `cmdx.createNode`, it works.

I'll look into adding an example and update the pr.

mottosso:
Hi @fvbehr! Any progress on this? I think it would be a neat feature to introduce that could help a lot of beginners. The remaining step is to fix the tests, and ideally add the same convenience to the remaining members, to avoid surprises.

fvbehr:
Hi @mottosso,
Sorry for the lack of updates. Had some tight deadlines. 
I am on vacation this week and will be able to look into it next week.

mottosso:
No problem, enjoy!

mottosso:
That looks good, could you add some examples in the docstrings too? Especially ones that capture what works and what doesn't.

mottosso:
Also have a look at what the [unitests say](https://mottosso.visualstudio.com/cmdx/_build/results?buildId=861&view=logs&jobId=aa1b1517-b9bf-50c8-ea98-af75e805f67c&j=aa1b1517-b9bf-50c8-ea98-af75e805f67c&t=0d1303e4-627b-5afb-e2a2-58a7a350376d)

mottosso:
Noticed you pushed another update, I think this looks good, nice work @fvbehr!

PR #23, opened 2020-10-01
Usability

mottosso:
Minor usability improvements.

- `cmdx.Plug.append(autofill)` for when you want to append to the next available index, rather than the largest index. That is, if an index is connected and then disconnected, the default index would be `1` even though `0` is available. Autofill attempts to utilise indices that are free by searching for the first unconnected index in an array attribute.
- Time plugs, e.g. `cmdx.encode("time1")["outTime"]`, now returns a value in `cmdx.Seconds`, which is consistent with how values are written to it, and to how they are written with the Maya Python API 2.0

### DIFF
--- a/cmdx.py
+++ b/cmdx.py
@@ -3247,7 +3247,7 @@ def _plug_to_python(plug, unit=None, context=None):
         6
         >>> "%.3f" % time["outTime"]  # Seconds
         '0.042'
-        >>> "%.2f" % time["outTime", UiUnit()]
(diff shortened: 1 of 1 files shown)


monkeez:
I noticed also that when working with `oma.MFnAnimCurve`, Maya sets and gets `MTime` in UI units. Would it be weird to use internal units for distance and angle, but UI units for time?

mottosso:
Let's stick to seconds. It makes no sense to default to UI units.

PR #24, opened 2020-10-05
Usability improvements

mottosso:
Lots of good stuff!


cmdx.py
```diff
@@ -2555,20 +2604,20 @@ def type(self):
 
         return self._mplug.attribute().apiTypeStr
 
-    def path(self):
-        return "%s.%s" % (
-            self._node.path(), self._mplug.partialName(
+    def path(self, full=True):
```
mottosso, line 2607: Interesting. Does this solve the issue of overly verbose paths e.g. `myNode.translate.translateX`? Or what was the motivation here?

cmdx.py
```diff
@@ -4096,9 +4192,38 @@ def createNode(self, type, name=None, parent=None):
         return DagNode(mobj, exists=False, modifier=self)
 
     @record_history
-    def parent(self, node, parent=None):
-        parent = parent._mobject if parent is not None else None
-        self._modifier.reparentNode(node._mobject, parent)
+    def parent(self, node, parent=None, preserve=True):
```
mottosso, line 4195: Need to document what's going on here.

cmdx.py
```diff
@@ -4096,9 +4192,38 @@ def createNode(self, type, name=None, parent=None):
...
+
+        if node.typeId == tJoint:
+            self.disconnect(node['inverseScale'])
+            if parent and parent.typeId == tJoint:
+                self.connect(parent['scale'], node['inverseScale'])
+
+    @record_history
+    def transform(self, node, transform):
+
+        if transform.rotationOrder() - 1 != node['ro']:
```
mottosso, line 4217: Need to document what's going on here. Why -1?

Also try and stick with double-quotes `"` instead of `'`, for consistency.

cmdx.py
```diff
@@ -4143,6 +4268,45 @@ def __exit__(self, exc_type, exc_value, tb):
 Context = DGContext
 
 
+def _encode_object(arg):
```
mottosso, line 4271: These need some documentation. Why do they exist, what do they do?

cmdx.py
```diff
@@ -5247,21 +5417,51 @@ def uninitializePlugin(plugin):
...
+tNonLinear = om.MTypeId(0x464e4c44)
 tNurbsCurve = om.MTypeId(0x4e435256)
 tNurbsSurface = om.MTypeId(0x4e535246)
-tJoint = om.MTypeId(0x4a4f494e)
+tOrientConstraint = om.MTypeId(0x444f5243)
+tPairBlend = om.MTypeId(0x4150424c)
+tPlusMinusAverage = om.MTypeId(0x52504d41)
+tPointConstraint = om.MTypeId(0x44505443)
+tPointOnCurveInfo = om.MTypeId(0x4e504349)
+tPointOnSurfaceInfo = om.MTypeId(0x4e505349)
```
mottosso, line 5456: Hm, I'm not convinced we need to add this many types to the module, when there is the option to use `cmdt` alongside `cmdx`. What do you think?

mottosso (commented):
Fantastic work @wougzy, have a look at my notes and let me know what you think.

cmdx.py
```diff
@@ -218,6 +218,8 @@ class _Type(int):
 kTransform = _Type(om.MFn.kTransform)
 kJoint = _Type(om.MFn.kJoint)
 kSet = _Type(om.MFn.kSet)
+kDeformer = _Type(om.MFn.kGeometryFilt)
+kConstraint = _Type(om.MFn.kConstraint)
```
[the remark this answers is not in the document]
wougzy: honestly the main reason i added them is because it was useful when i was refactoring my existing code (:

cmdx.py
```diff
@@ -732,12 +734,27 @@ def isA(self, type):
 
         return self._mobject.hasFn(type)
 
+    def isType(self, type):
```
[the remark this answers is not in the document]
wougzy: well if it doesn't lead to performance issues why not? I found it confusing at first to have different ways to check types between typeId, typeName, Fn.. especially when you need to pass them in a single argument. I wrote it to simplify this part of the connections methods that was hard to read at first. i wondered if the initial goal was to gain performance.

there's another spot in the code i hesitated to simplify (the get children part) but i wasn't sure how you wanted to design it so i left it as it is. but i'd suggest you have a look there too

cmdx.py
```diff
@@ -1062,17 +1082,32 @@ def connections(self, type=None, unit=None, plugs=False):
...
 
         for plug in self._fn.getConnections():
-            mobject = plug.node()
-            node = Node(mobject)
-
-            if not type or type == node._fn.typeName:
-                plug = Plug(node, plug, unit)
-                for connection in plug.connections(plugs=plugs):
+            plug = Plug(self, plug, unit)
+            for connection in plug.connections(type=type, unit=unit, plugs=plugs, source=source, destination=destination):
```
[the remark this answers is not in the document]
wougzy: sorry (: it's part of my bad coding habits that i have to change

cmdx.py
```diff
@@ -1062,17 +1082,32 @@ def connections(self, type=None, unit=None, plugs=False):
...
+                else:
                     yield connection
 
-    def connection(self, type=None, unit=None, plug=False):
+    def connection(self, type=None, unit=None, plug=False, source=True, destination=True, connection=False):
         """Singular version of :func:`connections()`"""
-        return next(self.connections(type, unit, plug), None)
+        return next(self.connections(type, unit, plug, source, destination, connection), None)
+
+    def inputs(self, type=None, unit=None, plugs=False, connections=False):
```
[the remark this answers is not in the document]
wougzy: my main concern was to find again the same way to code connections from pymel, the connections argument was also missing from the original cmds commands.

most importantly i found some issues and discrepancies between the methods from Node and Plug. now they are the same and they should work correctly

cmdx.py
```diff
@@ -2231,6 +2267,12 @@ def plug(self):
     def isArray(self):
         return self._mplug.isArray
 
+    @property
+    def arrayIndices(self):
```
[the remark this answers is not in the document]
wougzy: yup bad coding habits again i admit it

cmdx.py
wougzy: oh right! i've seen you were planning to write the relative/absolute in another part of the code (def parent from root i remember) you really need that for rigging so i added that in the modifiers. i'm not sure if it's the best design and if we have to move that to the root command. you tell me

cmdx.py
wougzy: that's it. but it was part of my first findings when i began to refactor existing code. maybe we should have more options to translate plug names like cmds (like shortnames)

i also wondered why you were using full path and not shortest path when you convert node and plugs to string. is it for performance issues? if not it would be better to use shortest path to find back the original maya behaviour to print node names. but again i'm not really sure of the initial intention of it

cmdx.py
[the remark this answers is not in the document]
mottosso: Aha, I see what you want it to do. Yes, let's move this to `isA`, I can see how it helps readability, and I wouldn't worry about the performance for making comparisons. When in doubt, profile it.

cmdx.py
mottosso: Generally, I would leave convenience functionality to cmds (or PyMEL). There's no end to how much of cmds we could reimplement and it would not only bloat cmdx but also introduce unneccesary bugs for little to no gain.

I would move this into your own codebase, as a wrapper around this basic call.

mottosso:
Also don't forget to have a look at the broken test. Looks like there's just one, related to how paths are returned.

cmdx.py
mottosso: I've found that paths are rarely useful with cmdx, since nodes are uniquely identified via their hash. And when I do want to know the path, I found it useful to mimic what filesystems do.

```c
c:\my\full\path

# Versus
full\path
```

Especially since the shortest path in Maya changes based on what else is around. It reads to me like an optimisation for serialisation (like the shortName of attributes) rather than anything someone would want to read.

In this case, `.translate.translateX` is *not* what one would expect to read, so I think we're moving in the right direction.

cmdx.py
wougzy: hmm right! i understand your point. feel free to drop it if you're not sure about it

cmdx.py
wougzy: ah yes sorry i forgot that one. it is simply a mapping between the rotateOrder plug from transform nodes and the rotationOrder from transformationMatrix

not very convenient.. (:

cmdx.py
wougzy: yeah this is more like my personal use of cmdx. feel free to drop it too

i found it useful to have these ones when i refactored scripts since they use a lot of maya commands and i had to constantly switch between strings and Node instances.

cmdx.py
wougzy: oh yeah no problem. i missed that one

so why don't you completely remove these ones from cmdx? you could just import them all from cmdt then

wougzy:
ok! i'll have a look this evening. that's the part where i'm not familiar with all of this (:

mottosso:
No problem, I'll guide you through it. :)

| Step 1 | Step 2
|:---|:---
| image | image
| Step 3 | Step 4
| image | image

Yeah, the UI/UX of that is terrible. xD

You can also run the test locally, which I would recommend. You'll just need a few things.

```bash
mayapy
nose
flaky
```

And then call..

```bash
cd cmdx/
mayapy run_tests.py
```

cmdx.py
mottosso: Because I'd like cmdx to be standalone for most people, but extensible for the power users. Some types, like `tTransform` is more common than `tPairBlend` for example. It's not written in stone.. maybe one day we'll either merge all of them into `cmdx.py` or make a Python package instead.

cmdx.py
```diff
@@ -2541,6 +2586,10 @@ def show(self):
 
         self.channelBox = True
 
+    @property
+    def editable(self):
+        return not self._mplug.isFreeToChange()
```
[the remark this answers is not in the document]
wougzy: yeah this one is important because you can edit animated plug and they are connected. you may need both

cmdx.py
wougzy: no problem. i'm revising my wrapper anyway so i'll drop all this extra code
i agree on the point everyone should be able to implement this lib the way they want

wougzy:
i'm a bit confused about the unit tests. it looks like they failed over already existing part of cmdx i didn't touched and that are quiet recent (the doctests failed at the docstrings of Plug.append and it tried to compare traceback log from failed on purpose commands)

otherwise i commited some cleanup and i removed my personal stuff that i moved to my wrapper

mottosso:
I had a look at the failing tests, this one:

```bash
======================================================================
FAIL: Doctest: cmdx.Plug.append
  File "C:\program files\autodesk\maya2020\bin\python27.zip\doctest.py", line 2226, in runTest
    raise self.failureException(self.format_failure(new.getvalue()))
AssertionError: Failed doctest test for cmdx.Plug.append
  File "C:\github\mottosso\cmdx\cmdx.py", line 2350, in append

File "C:\github\mottosso\cmdx\cmdx.py", line 2366, in cmdx.Plug.append
Failed example:
    node["notArray"].append(2.0)
Expected:
    ...
    TypeError: "|appendTest.notArray" was not an array attribute
Got:
      File "C:\program files\autodesk\maya2020\bin\python27.zip\doctest.py", line 1315, in __run
        compileflags, 1) in test.globs
      File "<doctest cmdx.Plug.append[5]>", line 1, in <module>
        node["notArray"].append(2.0)
      File "C:\github\mottosso\cmdx\cmdx.py", line 2390, in append
        raise TypeError("\"%s\" was not an array attribute" % self.path())
    TypeError: "appendTest.notArray" was not an array attribute
```

And it's subtle, but the problem is:

```py
# Expected
    TypeError: "|appendTest.notArray" was not an array attribute

# Got
    TypeError: "appendTest.notArray" was not an array attribute
```

Notice the missing `|`? It has to do with how attribute paths have changed from full to shortest. We should definitely keep showing the full paths, but I also noticed that..

```py
>>> persp = encode("persp")
>>> persp["translate"].path()
persp.translate
>>> persp["translateX"].path()
persp.translate.translateX
```

Where the `shortestPath` didn't address the issue of overly verbose attribute names, unless we pass `full=False`.

```py
>>> persp["translateX"].path(full=False)
persp.translateX
```

I think False should be the default, I can't see why one would ever want that extra level for attribute paths, it especially doesn't add to readability. Unless I'm overlooking something?

mottosso:
I made some changes but wasn't able to push to your repo, here's a patch it would be great if you could integrate.


<details><summary>0001-Reduce-default-verbosity-of-attribute-names.patch</summary>

```bash
From 025004bd2a86edded4ac9e244c06ec4055aa128c Mon Sep 17 00:00:00 2001
From: Marcus Ottosson <konstruktion@gmail.com>
Date: Wed, 7 Oct 2020 07:09:06 +0100
Subject: [PATCH] Reduce default verbosity of attribute names E.g. from
 .translate.translateX -> .translateX

---
 cmdx.py | 46 ++++++++++++++++++++++++++++++++++++++--------
 1 file changed, 38 insertions(+), 8 deletions(-)

diff --git a/cmdx.py b/cmdx.py
index 1f1006a..ef6c91b 100644
--- a/cmdx.py
+++ b/cmdx.py
@@ -2096,7 +2096,7 @@ class Plug(object):
         if isinstance(other, str):
             try:
                 # E.g. node["t"] + "x"
-                return self._node[self.name() + other]
+                return self._node[self.name(long=False) + other]
             except ExistError:
                 # E.g. node["translate"] + "X"
                 return self._node[self.name(long=True) + other]
@@ -2662,20 +2662,50 @@ class Plug(object):
 
         return self._mplug.attribute().apiTypeStr
 
-    def path(self, full=True):
+    def path(self, full=False):
+        """Return path to attribute, including node path
+    
+        Examples:
+            >>> persp = encode("persp")
+            >>> persp["translate"].path()
+            '|persp.translate'
+            >>> persp["translateX"].path()
+            '|persp.translateX'
+
+        """
+
         return "{}.{}".format(
-            self._node.shortestPath(), self._mplug.partialName(
+            self._node.path(), self._mplug.partialName(
                 includeNodeName=False,
                 useLongNames=True,
                 useFullAttributePath=full
             )
         )
 
-    def name(self, long=False, full=True):
-        return self._mplug.partialName(
-            includeNodeName=False,
-            useLongNames=long,
-            useFullAttributePath=full
+    def name(self, long=True, full=False):
+        """Return name part of an attribute
+
+        Examples:
+            >>> persp = encode("persp")
+            >>> persp["translateX"].name()
+            'translateX'
+            >>> persp["tx"].name()
+            'translateX'
+            >>> persp["tx"].name(long=False)
+            'tx'
+            >>> persp["tx"].name(full=True)
+            'translate.translateX'
+            >>> persp["tx"].name(long=False, full=True)
+            't.tx'
+
+        """
+
+        return "{}".format(
+            self._mplug.partialName(
+                includeNodeName=False,
+                useLongNames=long,
+                useFullAttributePath=full
+            )
         )
 
     def read(self, unit=None, time=None):
-- 
2.28.0.windows.1

```

</details>

Save that file, and then:

```bash
$ cd cmdx
$ git apply 0001-Reduce-default-verbosity-of-attribute-names.patch
```

cmdx.py
```diff
@@ -729,15 +734,26 @@ def isA(self, type):
...
+        if isinstance(type, om.MTypeId):
+            return type == self._fn.typeId
+        elif isinstance(type, string_types):
+            return type == self._fn.typeName
+        elif isinstance(type, (tuple, list)):
+            return self._fn.typeName in type or self._fn.typeId in type
+        elif isinstance(type, int):
+            return self._mobject.hasFn(type)
 
-        return self._mobject.hasFn(type)
```
mottosso, line 733: We should warn here, the user might not understand types don't match because of an invalid type.

```py
else:
  cmds.warning("Unsupported argument passed to isA('%s')" % type)
  return False
```

cmdx.py
```diff
@@ -1136,7 +1231,8 @@ def __repr__(self):
     def __init__(self, mobject, *args, **kwargs):
         super(DagNode, self).__init__(mobject, *args, **kwargs)
 
-        self._tfn = om.MFnTransform(mobject)
+        dagpath = om.MDagPath.getAPathTo(mobject)
+        self._tfn = om.MFnTransform(dagpath)
```
mottosso, line 1235: What's happening here? Why didn't passing `mobject` suffice?

cmdx.py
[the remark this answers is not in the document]
mottosso: We should make this more explicit.

```py
return self._mplug.isFreeToChange() == om.MPlug.kFreeToChange
```

Got a little confused at first as to why the `not` is there, and it just so happens that `kFreeToChange` is `0`. Poor choice of value. xD

wougzy:
omg i completely missed that "pipe" character in the logs. i forgot i modified the node path part of this function and i thought it was safe. a brilliant sample of why unit test is a good practice..

cmdx.py
wougzy: this one is also important if you want to use MSpace.kWorld with MFnTransform functions (like translation or rotatePivot for example)
it is also needed for MFnMesh getPoints with kWorld and so on..

mottosso: Hm I see.

I would be interested in the performance impact of this, since it would affect every instantiation of nodes throughout cmdx. Could you try running a simple loop with before/after?

Have a look at [plot.py](https://github.com/mottosso/cmdx/blob/e17862ac4a5814dbed1165c839c27067338146cd/plot.py#L155) for some examples of how you can profile things like this.

If it turns out to have a significant impact, we should make it optional (somehow) since I have yet to encounter a need for these missing features and would prefer not to have to pay for it until I do.

wougzy: i agree with that. i was not comfortable with this addition. i had the feeling it could address some performance issue. i'll try to profile that from your example

a solution would be to add a function to DagNode to convert its MFnTransform

wougzy: hmm the performance loss is significant
i ran the Test of createNode transform
i have on average 416ms for cmds, 429ms for cmdx and 466ms with this modification

mottosso: I can think of two options that would suffice.

1. Lazy instantiation, for example in the `translation()`, if `kWorld` is passed, instantiate it *once* there, and then reuse it the next time around.
2. Make free functions for e.g. `cmdx.translation(node, cmdx.sWorld)` that support the features you're missing.

The latter means you pay for it only for the nodes that matter, and only when it matters, only for those that actually use it. The former means the cost is explicit. Either option seems acceptable to me.

mottosso:
2020.3 was released yesterday with an interesting addition `MDGModifier.deleteNode(includeParents=True)`, I wonder if this could address the issues you were seeing? :O

- https://help.autodesk.com/view/MAYAUL/2020/ENU/?guid=__py_ref_class_open_maya_1_1_m_d_g_modifier_html

Not much additional info about why it was added.

cmdx.py
wougzy: i like it this way and it will justify the "re-implementation" of those functions to DagNode

wougzy:
haha thanks autodesk...
it looks like this is precisely the problem i had with the modifier

when i run this sample on maya 2018. everything is deleted...
`parent = mx.create_node(mx.tTransform)
child = mx.create_node(mx.tTransform, parent=parent)

with mx.DagModifier() as md:
  md.delete(child)
`
it also explains why it is so difficult to delete shapes

wougzy:
while we're at it, what do you think about updating DagNode default string representations?

```
    def __str__(self):
        return self.shortestPath()

    def __repr__(self):
        return self.shortestPath()
```

the point is to stick to how maya print nodes by default. i don't know if it has a performance impact. this is something i modified in my wrapper anyway and which is a concern for me as i use this behaviour a lot to verbose data. I have a very biased vision on the subject because we often work with very deep hierarchies in rig and you see what i mean this is hard to read (:

mottosso:
> while we're at it, what do you think about updating DagNode default string representations?

Hm, yes I can see what you mean. I would leave `__repr__` with the full path, but use the short path for `__str__`. Would that work for you?

wougzy:
yes that's it! i believe that's what most tds would expect as it's the default Maya behaviour. I'll see that tomorrow quietly, it looks like there is a lot unit test that depend on it

speaking about that i may have a last small question about default stuffs. I noticed the default keyable state of new created attributes was True. It is False by default in mel and cmds. I couldn't find what was the default setting of the api. So was it done on purpose?

mottosso:
> I noticed the default keyable state of new created attributes was True

You're right, that sounds inconsistent with the API as well.

- http://help.autodesk.com/cloudhelp/2020/ENU/Maya-SDK-MERGED/cpp_ref/class_m_fn_attribute.html#a9e68a8b4af016b37f6567cfa6d68e551

Happy to switch that default to False.

mottosso:
Let me know when you're happy @wougzy and we'll get this merged. If something isn't quite perfect yet, let's save it for another PR where we can focus in on just that, before this PR gets too big.

wougzy:
haha right. i think i'm done with theses ones yeah. i'll leave the str changes. i don't want to interfere with the unit test and i can deal with that in wrapper anyway

mottosso:
Whop! Before this PR gets too big, let's make it happen!

mottosso:
🥳

PR #25, opened 2020-10-06
DGContext uses Seconds by default, added UiUnits for each unit type

monkeez:
- Added `AngleUiUnit()` and `DistanceUiUnit()`
- Renamed `UiUnit()` to `TimeUiUnit()` for consistency
- `DGContext` now uses seconds by default
- `DGContext` supports a unit argument for different time units
```python
with cmdx.DGContext(1, cmdx.Minutes):
    pass
# Or
with cmdx.DGContext(cmdx.Minutes(1)):
    pass
```

### DIFF
--- a/cmdx.py
+++ b/cmdx.py
@@ -4116,7 +4116,7 @@ class DagModifier(_BaseModifier):
 
 class DGContext(om.MDGContext):
 
-    def __init__(self, time=None):
+    def __init__(self, time=None, unit=None):
         """Context for evaluating the Maya DG
 
         Extension of MDGContext to also accept time as a float. In Maya 2018
@@ -4128,8 +4128,9 @@ class DGContext(om.MDGContext):
         """
 
         if time is not None:
-            if isinstance(time, (int, float)):
-                time = Seconds(time)
+            if not isinstance(time, TimeType):
+                unit = unit or Seconds
+                time = unit(time)
             super(DGContext, self).__init__(time)
         else:
             super(DGContext, self).__init__()
--- a/cmdx.py
+++ b/cmdx.py
@@ -16,7 +16,7 @@ from maya import cmds
 from maya.api import OpenMaya as om, OpenMayaAnim as oma, OpenMayaUI as omui
 from maya import OpenMaya as om1, OpenMayaMPx as ompx1, OpenMayaUI as omui1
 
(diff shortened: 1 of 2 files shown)


mottosso:
Neat, but why does the context need a unit? :O

monkeez:
In case you want to evaluate in a time unit other than seconds. Should I rename the argument to timeUnit?

mottosso:
Ah, yes of course. That makes sense.

```py
with cmdx.DGContext(1, cmdx.Minutes):
    pass
```

I think this reads the most clear.

> Renamed UiUnit() to TimeUiUnit() for consistency

Better add an alias `UiUnit = TimeUiUnit` to not break existing code.

Other than that, I think this looks good!

mottosso:
The other syntax, `DGContext(cmdx.Minutes(1))` also works it looks like? I think that would be great as well, in case the value is being passed from some other function or is already an MTime.

monkeez:
That's correct, `DGContext(cmdx.Minutes(1))` should already work.

UiUnit alias added.

mottosso:
Loving it.

Can I hand you permissions to merge and make releases for cmdx? It's a little tedious, but it'd help me out a lot. This time around, it would involve:

1. Incrementing the version, based on https://semver.org/ so a patch version should do
2. Drafting a release [here](https://github.com/mottosso/cmdx/releases), and more or less copy/paste what you've already summarised in this PR, good job. The goal is being able to flick through the releases and understand what's been added and  changed, bearing in mind that we must never break backwards compatibility. Ever. Unless we increment the major version.
3. Done

A release will be made to PyPI automatically from there.

mottosso:
Gotta run, but you can go ahead when ready. If not, no worries and I'll return to this later today.

monkeez:
Awesome, thanks. I'll take a look at this later today (Sydney time 😄). I might have some more questions about doing a release.

monkeez:
Do you still want me to do a release or should we wait for #24?

mottosso:
Let's make a release for every PR, keeps it bite-sized and without doubt about when to make releases. Please go ahead! :)

mottosso:
https://github.com/mottosso/cmdx/releases/tag/0.4.7

Great job. :)

PR #29, opened 2020-10-17
Implement support for MColor and associated attributes

mottosso:
You can now do this.

```py
import cmdx
lambert = cmdx.encode("lambert1")
lambert["color"] = cmdx.ColorType(1, 0, 0)

# And this
color = cmdx.ColorType()
color.setColor((360, 0.5, 0.75), cmdx.ColorType.kHSV, cmdx.ColorType.kFloat)
lambert["color"] = color

# And this
from maya.api import MColor
lambert["color"] = MColor(0, 1, 0)
```

### DIFF
--- a/cmdx.py
+++ b/cmdx.py
@@ -94,6 +94,7 @@ MAngle = om.MAngle
 TimeType = om.MTime
 DistanceType = om.MDistance
 AngleType = om.MAngle
+ColorType = om.MColor
 
 ExistError = type("ExistError", (RuntimeError,), {})
 DoNothing = None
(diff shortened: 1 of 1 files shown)


cmdx.py
```diff
@@ -3753,6 +3754,11 @@ def _python_to_plug(value, plug):
         for index, value in enumerate(value):
             _python_to_plug(value, plug[index])
 
+    elif isinstance(value, om.MColor):
```
mottosso, line 3757: Note to future readers, I initially tried the same approach as with MVector, but for whatever reason I kept getting these black-box Maya errors about it. Unrolling that loop like this however works just fine.

PR #33, opened 2020-10-26
Add Component and Deformer classes

wougzy:


cmdx.py
```diff
@@ -14,7 +14,8 @@
 
 from maya import cmds
 from maya.api import OpenMaya as om, OpenMayaAnim as oma, OpenMayaUI as omui
-from maya import OpenMaya as om1, OpenMayaMPx as ompx1, OpenMayaUI as omui1
+from maya import OpenMaya as om1, OpenMayaMPx as ompx1, OpenMayaUI as omui1, \
```
mottosso, line 17: Death to backslashes.

```py
from maya import (
    OpenMaya as om1,
    OpenMayaMPx as ompx1,
    OpenMayaUI as omui1,
    OpenMayaAnim as oma1
)

File: cmdx.py [hunk 2]
```diff
@@ -357,11 +358,16 @@ def __call__(cls, mobject, exists=True, modifier=None):
         # It didn't exist, let's create one
         # But first, make sure we instantiate the right type
         if mobject.hasFn(om.MFn.kDagNode):
-            sup = DagNode
+            if mobject.hasFn(om.MFn.kMesh):
+                sup = Mesh
+            else:
+                sup = DagNode
```
mottosso, line 364: Could we flatten this if branch?

```py
        if mobject.hasFn(om.MFn.kMesh):
            sup = Mesh
        elif mobject.hasFn(om.MFn.kDagNode):
            sup = DagNode
        elif mobject.hasFn(om.MFn.kSet):
            sup = ObjectSet
        ...
```

cmdx.py
```diff
@@ -1246,10 +1252,16 @@ def __repr__(self):
...
 
-        # Convert self._tfn to om.MFnTransform(self.dagPath())
-        # if you want to use its functions which require sWorld
         self._tfn = om.MFnTransform(mobject)
 
+        # additional cache placeholders
+        self._mdagpath = None
+
+        # Create self._dagtfn with om.MFnTransform(self.dagPath())
+        # if you want to use dag node functions which require sWorld
```
mottosso, line 1261: - Lowercase sentence, `Additional...`
- Typo "Create self._dagfn..."

cmdx.py
```diff
@@ -1282,7 +1294,9 @@ def dagPath(self):
 
         """
 
-        return om.MDagPath.getAPathTo(self._mobject)
+        if self._mdagpath is None:
+            self._mdagpath = om.MDagPath.getAPathTo(self._mobject)
```
mottosso, line 1298: Are we sure this won't break, if the node is later moved? The node reference to cmdx would be reused, but this path might break. We might need to compute this every time.

cmdx.py
```diff
@@ -1781,23 +1795,83 @@ def setLimit(self, typ, value):
...
+        """Return the full vertex Component object of this mesh
+        """
+        if self._vtx is None:
+            k = om.MFn.kMeshVertComponent
+
+            cps = om.MFnSingleIndexedComponent().create(k)
+            fn = om.MFnSingleIndexedComponent(cps)
+            fn.setCompleteData(self.numVertices)
+
+            self._vtx = MeshVertex(self, cps)
```
mottosso, line 1846: Will this break, if new vertices are added to a mesh after first reference by cmdx? E.g. (1) `mesh.vtx` is accessed, (2) mesh is subdivided, (3) `mesh.vtx` is accessed again, (4) boom?

mottosso: Let's add some docstring examples here to test this.

cmdx.py
```diff
@@ -1781,23 +1795,83 @@ def setLimit(self, typ, value):
...
+
+        # update length if it changed
+        numVertices = self.numVertices
+        if len(self._vtx) != numVertices:
+            self._vtx._fn.setCompleteData(numVertices)
+
+        return self._vtx
+
+    def getPoints(self, space=sObject):
```
mottosso, line 1855: Missing docstring, and examples.

cmdx.py
```diff
@@ -1898,42 +1981,18 @@ def flatten(self, type=None):
...
-                elif type is not None:
-                    if type == member.typeName:
-                        members.add(member)
-                else:
-                    members.add(member)
-
-        recurse(self)
-
-        return list(members)
+        return iter(self.members(type=type, flatten=True))
```
mottosso, line 1984: This changes the return type from list to iterator, better stick with list here, especially since we aren't gaining anything from it being an iterator.

cmdx.py
```diff
@@ -2684,6 +2768,9 @@ def type(self):
 
         return self._mplug.attribute().apiTypeStr
 
+    # typeName is used as key for ObjectSet.sort
+    typeName = None
+
```
mottosso, line 2773: What's going on here? :O

cmdx.py
```diff
@@ -3861,13 +3948,144 @@ def _python_to_mod(value, plug, mod):
     return True
 
 
+class MetaComponent(type):
+
+    @withTiming()
+    def __call__(cls, mdagpath, mobject, exists=True, modifier=None):
+
+        if mobject.hasFn(om.MFn.kMeshVertComponent):
```
mottosso, line 3956: Can this not be a method of Component directly? Why the metaclass?

cmdx.py
```diff
@@ -3861,13 +3948,144 @@ def _python_to_mod(value, plug, mod):
...
+    path = __str__
+
+    def __nonzero__(self):
+        return not self._fn.isEmpty
+
+    __bool__ = __nonzero__
+
+    @property
+    def complete(self):
+        return self._fn.isComplete
```
mottosso, line 4005: Let's call this isComplete, otherwise it might be interpreted as a verb.

cmdx.py
```diff
@@ -3861,13 +3948,144 @@ def _python_to_mod(value, plug, mod):
...
+        if isinstance(index, slice):
+            fn.addElements(range(*index.indices(n)))
+        elif isinstance(index, tuple):
+            fn.addElements(index)
+        else:
+            if index < 0:
+                index = n + 1 + index
+            fn.addElement(index)
+
+        return self.__class__(self._node, cps)
```
mottosso, line 4043: Can we use `type(self)` instead of `__class__`? It's used elsewhere for this purpose.

cmdx.py
```diff
@@ -3861,13 +3948,144 @@ def _python_to_mod(value, plug, mod):
...
+            for i in self._fn.getElements():
+                yield i
+
+    def __getitem__(self, index):
+        cps = self._Fn().create(self._type)
+        fn = self._Fn(cps)
+        n = self.maxLength()
+        if n is None:
+            raise IndexError("{} doesn't support get item".format(self.__class__))
+
```
mottosso, line 4033: Missing docstrings and examples.

cmdx.py
```diff
@@ -3861,13 +3948,144 @@ def _python_to_mod(value, plug, mod):
...
+        for i in self:
+            it.setIndex(i)
+            yield i
+
+    def append(self, item):
+        if not isinstance(item, int):
+            raise TypeError()
+        if item >= self.maxLength():
+            raise IndexError()
```
mottosso, line 4057: Don't forget to include a message with the exception.

cmdx.py
```diff
@@ -3861,13 +3948,144 @@ def _python_to_mod(value, plug, mod):
...
+            self._fn.addElement(item)
+
+    @property
+    def elements(self):
+        return self._fn.getElements()
+
+
+class MeshVertex(Component1D):
+    _type = om.MFn.kMeshVertComponent
+
```
mottosso, line 4068: Can we make `MeshVertex` and `Component1D` private? Do they need to be accessed independently? Let's call these `_MeshVertex` and `_Component1D` to let the user know these are internal-only.

cmdx.py
```diff
@@ -3881,9 +4099,74 @@ def encode(path):
         raise ExistError("'%s' does not exist" % path)
 
     mobj = selectionList.getDependNode(0)
+    if "." in path:  # costs 0.2µs
```
mottosso, line 4102: Now we need to include tests for all combinations of paths passed to encode.

cmdx.py
```diff
@@ -3881,9 +4099,74 @@ def encode(path):
...
+    """Convert relative or absolute `pathList` to cmdx objects
+
+    Fastest conversion from absolute path to Node/Plug/Component
+
+    Arguments:
+        pathList (list): Absolute or relative path to DAG/DG nodes,
+            plugs or components
+
+    """
+    assert isinstance(pathList, (list, tuple)), "%s is not a list" % pathList
```
mottosso, line 4126: Need docstring tests/examples.

cmdx.py
```diff
@@ -3881,9 +4099,74 @@ def encode(path):
...
+    for encoded in encodeSelectionList(selectionList):
+        yield encoded
+
+
+def encodeSelectionList(selectionList):
+    """Convert a native MSelectionList to cmdx objects
+
+    Arguments:
+        selectionList (om.MSelectionList)
+    """
```
mottosso, line 4145: Missing examples.

cmdx.py
```diff
@@ -4004,6 +4290,19 @@ def _encode1(path):
...
+        try:
+            mdag = om1.MDagPath()
+            cp = om1.MObject()
+            selectionList.getDagPath(0, mdag, cp)
+            return (mdag, cp)
+        except:
+            mplug = om1.MPlug()
+            mplug = selectionList.getPlug(0, mplug)
+            return mplug
+
```
mottosso, line 4305: Did this need to change? It's only used internally, is anything using the added features?

mottosso (changes requested):
Whop!

wougzy:
hehe of adjustments ahead (: i'll comments all of these tonight. we're about to get a tight schedule at studio so i won't have too much time in the next days

cmdx.py
wougzy: yeah alright! i never questionned that as that's something pycharm would wrap automatically. but i didn't liked it so i'm happy to know that these "\" are ugly

cmdx.py
wougzy: I agree for the readabilty. I did this to gain some µs from the MFn check of shape which is useless once you did the one for DagNode. The point was to reproduce the inheritance of classes there

cmdx.py
mottosso: Death to PyCharm too. ;)

cmdx.py
wougzy: i have to pay attention to lowercase. I'm used to lowercase everthing. bad habit etc.. (:

and for self._dagtfn it was not a typo but i'll rewrite some additional explanations for these. Maybe i'll revise these cache from what you noticed with dagpath cache

cmdx.py
wougzy: I tested that behaviour it looks solid. I changed the hierarchy of these and everything still works fine (from str to matrices)

Caching MDagPath instances have huge benefits for castring components. I'll try some other tests just to be sure

cmdx.py
wougzy: I added a sync somewhere in the code for this purpose. so the _fn of the components would work even if the geometry changed.

But you're right, I'll add some docstring tests!

cmdx.py
wougzy: yeah ok! i didn't paid attention to this. It suffered from the legacy translating. And besides, I added a flatten flag directly to .members() method which is alrerady an iterator

cmdx.py
wougzy: When i tried to use ObjectSet.sort() it crashed when i had components or plugs in it. So i had to had a typeName variable for them just to make the .sort work

This is probably not the best approach. I'll have to rewrite .sort to avoid adding that

cmdx.py
wougzy: Yeah i was unsure about it. I did this to copy how it was implemented on Node. But this was lazy, I think it can be rewritten as you propose

cmdx.py
wougzy: alright!

cmdx.py
wougzy: yep you're right! there is no point to cast Component1D and MeshVertex directly

cmdx.py
wougzy: Oh no it's a scar from a previous implementation test of component with legacy support of ObjectSet

I'll rollback this. Especially if we try to avoid legacy in the future..

cmdx.py
```diff
@@ -1931,13 +1931,18 @@ def clear(self):
...
 
         Arguments:
             key (lambda): See built-in `sorted(key)` for reference
 
         """
+        if key is None:
+            key = lambda o: (
+                o.typeName if isinstance(o, Node) else None,
+                o.path()
+            )
```
mottosso, line 1946: Could save on indentation levels and remove the if-statement, by:

```py
            key = key or lambda o: (
                o.typeName if isinstance(o, Node) else None,
                o.path()
            )
```

cmdx.py
mottosso: I suspect the intended usage of this and other functions sets is:

```py
fn = om.MFnSingleIndexedComponent()
cps = fn.create(k)
```

Where `fn` would get the latest created MObject associated with it automatically.

wougzy: I investigated a bit on this one. there is nothing to optimize here. It looks like there fn.create is a sort of classmethod of MFnSingleComponent. caching an void instance of it has not impact on performance.

The way the implemented components in maya is a bit weird. A component is a combination of a node (MDagPath) and a index list with a specified type. So you don't have any direct link between these 2 objects. What i mean is if you want to be sure to not have any problem with getting indices out of range, we have to check that ourselves and that's what i added in the code. So reusing the same casted component MObject doesn't really matter (well that's how i understood it)

And yes I have to add these docstring check (: (and every other missing comments)

cmdx.py
wougzy: I made only _Component1D private. It should be possible to test instance type of casted Component with `isistance(cmdx.MeshVertex, object)` (or we have to implement something like `Node.isA`)

cmdx.py
wougzy: oh right! I used to be careful with the or operator. I had a few problems when comparing stuff with None and the result was 0 instead. But there is no chance this will be a problem here

mottosso:
It's tricky for me to see which comments have been addressed; are you able to hit "Resolve Conversation" once you've addressed it?

image

mottosso:
I can't remember where we left this, I remember we had a conversation outside of this PR about it? :S What's missing here is a motivation for adding this amount of code, along with some examples of how to use it and what it is for.

mottosso:
Hate to close something with this amount of work. :( But it really does need reducing; it's hard to spot and test this many changes, and backwards compatibility is of utmost importance. A new (smaller) PR is welcomed!

PR #36, opened 2020-11-04
small documentation corrections:

benblo:
- typos
- make DagModifier examples more consistent (there were name changes in-between examples, making it difficult to follow and compare)


mottosso:
Nice one, thanks @benblo

PR #37, opened 2020-11-04
support attribute assignment from an on.MVector

benblo:


mottosso:
I wonder if we need to add support to other instances of `(tuple, list)` as well then, to avoid surprises? E.g. line 521 for adding a new attribute.

benblo:
I'm not sure, tbh I don't really understand what's going on there... it's looking for a `(Type[_AbstractAttribute], dict)` pair?

One candidate I thought I saw was in `_python_to_plug` line 3753, but in fact it's handled above at 3708 (matrix type/length is handled explicitly, but vector isn't).

We could also declare a `vector_types = list, tuple, om.MVector` (like `string_types` does) to make it clearer what the `isinstance()` calls are looking for.

BTW the code uses sometimes `(list, tuple)`, sometimes `(tuple, list)`, which does not make it easy to look for all usages :). I wonder if something like `typing.Sequence` would be applicable here? (I'm still fuzzy on Python typing...)

mottosso:
That's true, it should really have some comments to break that down, took me a while to decrypt it.

But let's break it down.

```py
        # E.g. node["newAttr"] = cmdx.Double(default=5)
        # Double is a subclass of `dict`, but when called without arguments
        # returns a tuple of (AttributeType, {"default": 5, etc})
        elif isinstance(value, (tuple, list)):

            # Is it a class instance, something yet to be instantiated? E.g. cmdx.Double
            if isinstance(value[0], type):

                # Is it an attribute, or some other unknown class?
                if issubclass(value[0], _AbstractAttribute):

                    # Ok, it's one of ours, let's create it
                    Attribute, kwargs = value
                    attr = Attribute(key, **kwargs)
                    try:
                        return self.addAttr(attr.create())
                    except RuntimeError:
                        # NOTE: I can't be sure this is the only occasion
                        # where this exception is thrown. Stay catious.
                        raise ExistError(key)
```

The reason being that you can create an attribute in a few different ways..

```py
node.add_attr(cmdx.Double("newAttr"))
node["newAttr"] = cmdx.Double()
```

In the latter case, the name of the attribute comes from the caller, rather than the attribute instance itself.

benblo:
Ah ok, I would not have guessed that!
So in that case it's unrelated, isn't it? we're never gonna find an `MVector` in a `(type, dict)` pair. And, I assume the `MVector` case is handled in `attr.create()` by the normal python-to-plug mechanism.
It looks like my fix was limited to the unit-conversion case (which I don't think I realized at the time).

BTW I had noticed that `_AbstractAttribute` inherits from `dict`, is there a specific reason for this? versus storing stuff in an internal dict field, or even using python attributes...
It feels a bit weird to do eg `self["name"]` instead of `self.name`.

mottosso:
> BTW I had noticed that _AbstractAttribute inherits from dict, is there a specific reason for this?

It started as just a `{"name": "attr", "default": 5}` etc. and eventually grew into a class. I'd prefer keeping it a dict, to keep anyone from adding complexity to it, like `@property` etc. It's data, not behavior.

mottosso:
Let's get this one out of the way, thanks @benblo

Issue #5, opened 2019-11-11
Compound of Compounds


mottosso:
For some reason, compound children of a compound locks up Maya, presumably there's an infinite loop somewhere.

```py
import cmdx

node = cmdx.createNode("transform")

# This works
node.addAttr(
    cmdx.Compound("parent", children=[
        cmdx.Double("age"),
        cmdx.Double("height")
    ])
)

# This does *not*
node.addAttr(
    cmdx.Compound("parent", children=[
        cmdx.Compound("child", children=[
            cmdx.Double("age"),
            cmdx.Double("height")
        ])
    ])
)
```

benoit-fouletier:
I was just about to open an issue about this, this indeed hangs hard. Any luck finding a workaround?

mottosso:
No, sorry. I've avoided it ever since :( If you happen to come across one, please do share!

benblo:
OK, found the issue (and a fix)!

The issue is, in `Compound.create()` (and actually in every attribute type), you use the same global (static) function set for every instance. I suspect this is a problem, since the `Fn` must actually be associated with the actual `MObject` upon `create()` . So when later you do:
```
	self.Fn.addChild(child.create(cls))
```
the child can be added to the correct `mobj`... which means you can't reuse `Fn`, it has to be unique. The same might be true (?) for the other attribute types.

I saw examples online that make this clearer:
```
	fn = om.MFnDependencyNode()
	fn.setObject(mobj)
```
... I have to admit I don't understand the API's terminology there, if the function set is indeed associated with an object then it's not really a function set but an instance isn't it, so what's the difference? beats me.

Also, I noticed that you're using `self.Fn`, but `Fn` is actually declared as a class attribute, not an instance attribute, so shouldn't it be `Compound.Fn` instead? I don't mean to be pedantic, this usage might be pythonic, but I'm quite new to Python (and Maya too actually) so still getting used to some things that feel odd to me coming from C#/C++ :relaxed:.

---

Anyway, not sure I have the correct diagnosis, but to fix this I instantiated a new `Fn` for each instance:
<pre>
    def __init__(self, name, children=None, **kwargs):
<b>        self.Fn = om.MFnCompoundAttribute()</b>
        ...
</pre>
... no more hangs!
(I don't have time atm to make a proper fork and PR, but once I do (hopefully) I'll have a few other fixes to send your way.)

---

... that being said, unfortunately the attribute editor doesn't show nested compounds :( !
Here's a simple test:
```
[...]

PR #38, opened 2020-11-04
fix hang when adding nested compound attributes

benblo:
fixes https://github.com/mottosso/cmdx/issues/5

### DIFF
--- a/cmdx.py
+++ b/cmdx.py
@@ -5258,6 +5258,7 @@ class Compound(_AbstractAttribute):
     Multi = None
 
     def __init__(self, name, children=None, **kwargs):
+        self.Fn = om.MFnCompoundAttribute()  # see https://github.com/mottosso/cmdx/issues/5#issuecomment-717330454
         if not children and self.Multi:
             default = kwargs.pop("default", None)
             children, Type = self.Multi
--- a/cmdx.py
+++ b/cmdx.py
@@ -5254,11 +5254,44 @@ class Distance(AbstractUnit):
 
 
 class Compound(_AbstractAttribute):
+    """One or more nested attributes
+    
+    Examples:
+        >>> _ = cmds.file(new=True, force=True)
+        >>> node = createNode("transform")
+        >>> node["compoundAttr"] = Compound(children=[
+        ...     Double("child1", default=1.0),
+        ...     Double("child2", default=5.0)
+        ... ])
+        ...
+        >>> node["compoundAttr"]["child1"].read()
+        1.0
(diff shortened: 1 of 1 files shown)


cmdx.py
```diff
@@ -5258,6 +5258,7 @@ class Compound(_AbstractAttribute):
     Multi = None
 
     def __init__(self, name, children=None, **kwargs):
+        self.Fn = om.MFnCompoundAttribute()  # see https://github.com/mottosso/cmdx/issues/5#issuecomment-717330454
```
mottosso, line 5261: Should we make the class `Fn = None`? Of of these aren't necessary.

Also try and stick to 80 lines with if possible, which it can be here if we put the comment on the line above.

benblo: Good point, removed.

Actually I was wondering if we should keep declaring `Fn` as currently (static for each attribute type), but always instantiate something like `self.Fn = cls.Fn()` directly in `_AbstractAttribute.__init__()`, but I'm not 100% sure this is strictly necessary outside for non-Compound attributes.
Also I'm not sure how to actually get to the type-specific `Fn` from `_AbstractAttribute` 😄

mottosso:
Could we add just one test to this, to make sure it works on all Maya's, and that it won't break again in the future?

benblo:
I could, though last time I used tests wasn't very fun ;)
I don't really have time to spend on this right now, I tend to work in bursts so next time I can focus on cmdx, I'll do it right, setup the docker thing you mentioned etc and finish this.

Btw, I recently created wrappers for some APIs (eg [playbackOptions](https://help.autodesk.com/cloudhelp/2020/ENU/Maya-Tech-Docs/CommandsPython/playbackOptions.html)), that are exposed as commands although to me they are much better suited as a class with properties (all it does is expose get/set accessors).
Much of my pains with Maya come from not having proper autocompletion for cmds, not having proper typing information, and spending way too much time reading the docs to figure out the keyword combinations, so I've now taken the habit of whenever I need a new API, first wrapping it into something I find more sensible (I feel like most of it could be automated).
I don't know if this is of any interest to you? maybe in a different module, since I feel this is not necessarily the focus of cmdx.

mottosso:
> I could, though last time I used tests wasn't very fun ;)

Haha, I understand exactly. xD I've added a test just now.

> I don't know if this is of any interest to you? maybe in a different module, since I feel this is not necessarily the focus of cmdx.

Not for cmdx, that's more in the territory of PyMEL I think; i.e. all-encompassing mess. What's keeping you from using that?

PR #40, opened 2020-11-04
Node/DagNode: str constructor

benblo:
- allows `cmdx.Node(node_path)` syntax as an alternative to `cmdx.encode(node_path)`

### DIFF
--- a/cmdx.py
+++ b/cmdx.py
@@ -338,7 +338,11 @@ class Singleton(type):
     _instances = {}
 
     @withTiming()
-    def __call__(cls, mobject, exists=True, modifier=None):
+    def __call__(cls, mobject, exists=True, modifier=None):  # type: (Type, Union[om.MObject, str], bool, bool) -> Node
+        # string constructor -> use encode()
+        if isinstance(mobject, string_types):
+            return encode(mobject)
+
         handle = om.MObjectHandle(mobject)
         hsh = handle.hashCode()
         hx = "%x" % hsh


mottosso:
Hmmmmmmmmm. I'm weary of this change.

- Should we also support `Node("pCube1.translateX")`
- And what about `Node("pCubeShape1.vtx[3]")`?
- And what about `Node("translateX_animCurve")`?

I think `encode` is in a better position to deal with the conversion of paths into their corresponding node type.

benblo:
Tbh `encode()` really threw me off when I first started using cmdx, the name meant nothing to me, and the return value doesn't provide any typing information.
If I had eg the name of a transform, I knew it was a DagNode, so my first instinct was to type `node = DagNode(name)`.
With this syntax, I have explicit typing and autocompletion.
If you don't like having it in the constructor, we could have a factory method like `Node.from_str(name)` (not my first choice because constructor are more natural, but well).

mottosso:
I see your point.

Let's figure out a way to handle the cases above; constructors aren't expected to throw errors, and factory functions won't account for passing in the name of a Dag node to a DG node type. With this change, you could even pass a animcurve path into a Node, and not get the Node you asked for.

mottosso:
An alternative could perhaps be..

```py
cmdx.node("|path|to|any|node")  # -> cmdx.DagNode etc.
```

About typing information, how do you mean? It should return you a fully qualified instance of either `cmdx.Node` or `cmdx.DagNode`. Sounds like maybe you are using an IDE that found soemthing it doesn't like.

benblo:
> constructors aren't expected to throw errors

Why is that? aren't they like any other function?

> factory functions won't account for passing in the name of a Dag node to a DG node type

I imagined some generic kung-fu could get us there, but I haven't really thought it through.

> With this change, you could even pass a animcurve path into a Node, and not get the Node you asked for.

Indeed (although I would argue that AnimCurve does derive from Node 😄, but your argument stands if you do `DagNode("my_curve")`, that's a mismatch).
We could actually type-check that the MObject is of the expected type and assert (the more type safety the better IMO).

> About typing information, how do you mean?

Just to be clear, I'm talking about static typing/analysis, not debugging. And, I'm using VsCode.
According to VsCode, `cmdx.encode()` returns `Node`, so eg `.children()` is unknown:
image
I can cast to DagNode to get autocompletion (notice how `children` is now yellow instead of white):
image
... but this is a bit tedious so I figured I'd just cut the middle-man.

In general, VsCode is often quite confused about the argument and/or return types; it's using the Pylance analyzer for autocompletion, which I'm not sure supports docstring type info (if there is even such a thing, again, Python noob here).
Other than that, that IDE is a **blast** to use. I have no experience with other Python IDEs, so I don't know if I should expect more.

mottosso:
Let's go down the route of `cmdx.node`. That way, it's clear it's a function and that function can then perform the magic necessary to turn strings into `cmdx.Node` instances. I don't like having the constructor reach out to another function which calls the same constructor again. It's too spaghetti.

The simple solution would be..

```py
def encode(...):
  ...

node = encode
```

To establish an alias. Maybe that's a better name for that function to begin with.

benblo:
How about this?
``` python
    T = TypeVar("T", bound="Node")
    @classmethod
    def from_name(cls, name):  # type: (Type[T], str) -> T
        node = encode(name)
        assert isinstance(node, cls), "node '%s' is of type %s instead of the expected %s" % (name, type(node), cls)
        return node
```
Usage:
``` python
sphere = cmds.sphere()[0]
node = cmdx.Node.from_name(sphere)

# or better:
node = cmdx.DagNode.from_name(sphere)

# this will fail:
node = cmdx.AnimCurve.from_name(sphere)
# AssertionError: node 'nurbsSphere1' is of type <class 'cmdx.DagNode'> instead of the expected <class 'cmdx.AnimCurve'>
```
---
> I don't like having the constructor reach out to another function which calls the same constructor again. It's too spaghetti.

Agreed, although IMO you're looking at this the wrong way: `encode` _is_ a constructor (or at least a [factory method](https://en.wikipedia.org/wiki/Factory_method_pattern)), because that's precisely what it does: return a new object from an argument; it should not be a function to begin with.
So you could just as well move `encode`'s code in the Node ctor (or Singleton, I don't clearly remember what's the back-and-forth there... could be argued it's spaghetti to begin with : ), and `encode` would just become a wrapper for Node("str").

I'll reiterate my first experience with cmdx: "I want a ref to a node; I have its name; first thing I try: cmdx.Node('name')".

(I first started using Maya _with_ cmdx, I had almost no prior experience with Maya; I now realize that the entire cmds API is string-based, but this wasn't ingrained for me -and from working with more modern engines, pretty lame: strings are slow and evil.)

mottosso:
I understand where you're coming from. I just don't agree. :) Having `cmdx.node` seems sufficient to resolve the initial problem of discovery.

benblo:
But it does not solve the typing issue though. With the `from_name` factory, you do get correct typing back in the IDE, _and_ typing validation at runtime. I forgot to include a screenshot earlier:
image
Which means I get better autocomplete after assignation (all the members from DagNode that are not in Node).

mottosso:
That's.. true, but I don't see the problem? When would you want to try and instantiate an anim curve from the path of a sphere? Need example(s).

mottosso:
Dropping this for now, it needs more substance as to why it should exist.

PR #41, opened 2020-11-04
add some Pylance-friendly typing information

benblo:


cmdx.py
```diff
@@ -1,5 +1,16 @@
 # -*- coding: utf-8 -*-
 
+# Hack for static typing analysis:
+# the test below only passes in the IDE (eg VsCode); Maya doesn't know or care about typing.
+MYPY = False
```
mottosso, line 5: Can MYPY be called something else, like `_MYPY` or `_TYPINGSUPPORT`? Best not to expose members to `cmdx` unless we intend on others using it and us supporting it. Do we need it, or could it if be `if False:`?

benblo: No, it has to be MYPY, it's a weird hack where the analyzer looks for this specific token and forcibly considers it true: https://mypy.readthedocs.io/en/stable/common_issues.html#import-cycles
... I agree the naming is unfortunate.

mottosso: Ah, could we at least `del MYMY` either directly after the `if`, or towards the end of the file? To avoid anyone using `cmdx.MYPY`

benblo: Ah yes we can apparently, done.

mottosso:
In that case I'm happy. Except we need an explanation in the release notes about why this is here. As someone who's never used or heard of pylance, could you write something up to get people going? I'll paste that into the upcoming release notes, or better yet you are welcome to do that as well.

benblo:
Actually Mypy seems to be quite widespread, if I understood correctly it started as an external project that is now the basis of Python 3's built-in type hinting.
And, it can be used as an offline type checker/static analyzer, not just for autocomplete (ie Pylance is just one user of it).

Would this do for a release note? Happy to add it, where would it go?

> Added the [MYPY](http://mypy-lang.org/) [token](https://mypy.readthedocs.io/en/stable/common_issues.html#import-cycles) to avoid false positives in some static analyzers (eg Pylance from VsCode is Python3-only, so it doesn't know about Python2 types like `basestring`).
> Also imports [`typing`](https://docs.python.org/3/library/typing.html) to add support for type hinting (IDE-only as `typing` is not included in Maya's Python distribution).

mottosso:
Thanks @benblo!

PR #42, opened 2020-11-04
dump() option to preserve attribute order

benblo:
- in the spirit of https://www.python.org/dev/peps/pep-0520/
- fix `sort_keys` wasn't using snake_case


mottosso:
Nice!

PR #46, opened 2020-12-05
Fix disconnect/connect with Undo

### LINKED ISSUE #45: DGModifier.disconnect and Undo
For whatever reason, you can't disconnect and connect in the same "do_it" batch without breaking undo.

```py
tx = cmdx.createNode("animCurveTL")
tx.keys(times=[1, 2, 3], values=[0.0, 1.0, 0.0])

tm = cmdx.createNode("transform")
tm["tx"] << tx["output"]

  mod.disconnect(tm["tx"])
  mod.connect(tm["sx"], tm["tx"])

cmds.undo()
# Error: (kFailure): Unexpected Internal Failure
# Traceback (most recent call last):
#   File "C:/cmdx.py", line 5548, in undoIt
#     shared.undos[self.undo]()
# RuntimeError: (kFailure): Unexpected Internal Failure # 
```

mottosso:
This fixes #45 

I've also moved the `MYPY` statement to below the import statements to make linters happy; but I wonder whether this breaks things for you @benblo?

### DIFF
--- a/cmdx.py
+++ b/cmdx.py
@@ -4262,8 +4262,11 @@ class _BaseModifier(object):
 
         if force:
             # Disconnect any plug connected to `other`
-            for plug in dst.connectedTo(True, False):
-                self.disconnect(plug, dst)
+            self.disconnect(dst)
+
+            # Connecting after disconnecting breaks undo,
+            # unless we do it first.
+            self._modifier.doIt()
 
         self._modifier.connect(src, dst)
 
--- a/cmdx.py
+++ b/cmdx.py
@@ -4280,6 +4280,32 @@ class _BaseModifier(object):
 
     @record_history
     def connect(self, src, dst, force=True):
+        """Connect one attribute to another, with undo
+
+        Examples:
+            >>> tm = createNode("transform")
+            >>> with DagModifier() as mod:
+            ...   mod.connect(tm["rx"], tm["ry"])
+            ...
+            >>> tx = createNode("animCurveTL")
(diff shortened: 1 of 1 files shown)


mottosso:
This is still *somewhat* fragile; the user can still..

```py
mod.disconnect(a, b)
```

Which would try and disconnect both source and destination connections to both a and b, even if no connections exists. If that's the case, Maya will happily carry on. Until you undo, which is when it'll tell you "hey, something you did at some point in the past was wrong, so now I will throw an exception. Happy hunting."

We can address this by testing that an attribute is actually connected before attempting to disconnect it.. but that's somewhat of a rabbit hole. As what really matters is what is connected upon calling `doIt`. A connection could not exist when calling `disconnect` but then later be made, which would invalidate whatever check we perform during the call to `disconnect`.

So I'm leaving that as-is for now. Hasn't been a problem so far.

mottosso:
Took another crack at this, and oh man.

```py
tm = cmdx.createNode("transform")
tm["tx"] << tm["ty"]

  mod.disconnect(tm["ty"], tm["tx"])

# OK
```

Except, it's *not* OK. Maya cares deeply about the order in which you specift the disconnect. Here, it'll try and disconnect `tx` from `ty`, except `tx` is not connected to `ty`. It's the other way around. So guess what happens? Nothing, ignored. And guess what happens on redo? `ty` gets a connection to `tx`. xDDD Lol. That's bad.

PR #54, opened 2021-03-24
Add support for Maya 2022

mottosso:
Python 3 includes the name of the module in errors raised, so `ExistError` became `cmdx.ExistError`. That complicated the doctests somewhat, but otherwise nothing has changed functionally.


mottosso:
There we go, no compromise, Maya 2022 and Python 3 is ready to go!

Issue #53, opened 2021-03-20
Deploy on CI


mottosso:
@monkeez I've moved CI from Azure to GitHub Actions, since Azure wasn't playing ball anymore. It wasn't providing a good UI on GitHub either, making it hard to spot when and where things went wrong.

The documentation deployment got lost in the transition, and there are a few too many things I don't recognise in the setup. Would you happen to have a moment to see if you can port it over?

```yaml

  # -----------------------------------------------------------------------
  #
  # Deploy docs
  #
  # -----------------------------------------------------------------------

  - job: Deploy_Docs
    condition: startsWith(variables['Build.SourceBranch'], 'refs/tags')
    pool:
      vmImage: "ubuntu-latest"
    steps:
      - task: DownloadPipelineArtifact@2
        inputs:
          artifact: docs
          path: $(Build.SourcesDirectory)/build/html

      - script: |
          git init
          git config --local user.name "Azure Pipelines"
          git config --local user.email "mitchellcoote@gmail.com"
          git add .
          touch .nojekyll
          git add .nojekyll
          git commit -m "Commit generated documentation"
        displayName: "Commit pages"
        workingDirectory: build/html

      - task: InstallSSHKey@0
        inputs:
          knownHostsEntry: $(KNOWN_HOST)
          sshKeySecureFile: deploy_key

      - script: |
          git remote add origin git@github.com:mottosso/cmdx.git
          git push -f origin HEAD:gh-pages
        displayName: "Publish GitHub Pages"
        workingDirectory: build/html

    dependsOn: Maya
```

I think it's safe to write as a separate "workflow" that only triggers on tags, similar to the PyPI deployment workflow currently in place.

monkeez:
I'll will try find some time to look into it, might be a couple of weeks though.

PR #57, opened 2021-04-19
Autodeploy docs to gh-pages

monkeez:
I had to merge the GitHub workflows into one yml file to get the `docs` job to depend on the `maya` job. The benefit is now the `deploy` job also depends on the `maya` job so it'll only push releases to PyPI that have passed the tests.


monkeez:
Resolves #53

mottosso:
Nice!

> I had to merge the GitHub workflows into one yml file to get the docs job to depend on the maya job.

Why was that? I didn't think the docs needed the results of the build anyway?

> The benefit is now the deploy job also depends on the maya job so it'll only push releases to PyPI that have passed the tests.

I actually specifically made it so we could deploy without going through yet another build. Since deployment only happens from the master branch, and nothing is merged with master unless it builds, it just adds friction. And the few times when you actually do want to deploy with a failing build, now we can't. :(

So ideally, deploy would remain separate. Is that possible you think?

mottosso:
Ok, I see now. Docs are built with mayapy.. That makes sense.. In that case, this looks solid. Let's try it!

monkeez:
Deploy could still be separate if you like, you'd just have to add the `deploy.yml` back and remove the `deploy` job from `main.yml`. I'd imagine that would all still work OK.

mottosso:
That's true, deploy and docs could remain separate. Oh well, let's give this a try and see what problems if any come out of it. Thanks @monkeez!

PR #62, opened 2021-06-16
Fool-proof plug-in load

### LINKED ISSUE #61: Two users, one machine
To support undo, cmdx installs itself as a plug-in using a unique filename. However! The file is written using the currently logged on users permission bits and places it in the system-wide temp directory. Therefore, if another user logs in and attempts to use cmdx, bam!

This is a bug.

mottosso:
Fixes #61 

**Special Case Fix**

This also addresses a special case of querying connections to elements of an array.

```py
b["arrayAttr"] = Long(array=True)
b["arrayAttr"][0] >> a["ihi"]
assert a["ihi"].connection() == b
assert a["ihi"].connection(plug=True) == b["arrayAttr"][0]
```

This didn't use to work, as the plug was automatically converted from a networked to non-networked plug using `findPlug`. And it didn't like indices in the name given, e.g. `worldMatrix[0]`.

**strict=True**

Turns out, `MSelectionList.add()` is lazy. It'll accept the shortest path to a node, such as `myNode` even though the actual name is `someNamespace:|myNode`. So, does `myNode` exist? No! That's what strict does.

```py
assert cmdx.exist("myNode", strict=False)
assert not cmdx.exist("myNode", strict=True)
assert not cmdx.exist("myNode")  # New default
```

### DIFF
--- a/cmdx.py
+++ b/cmdx.py
@@ -19,7 +19,7 @@ from maya import cmds
 from maya.api import OpenMaya as om, OpenMayaAnim as oma, OpenMayaUI as omui
 from maya import OpenMaya as om1, OpenMayaMPx as ompx1, OpenMayaUI as omui1
 
-__version__ = "0.6.0"
+__version__ = "0.6.1"
 
 PY3 = sys.version_info[0] == 3
 
--- a/cmdx.py
+++ b/cmdx.py
@@ -2652,13 +2652,13 @@ class Plug(object):
             >>> plug[1] << tm["parentMatrix"][0]
             >>> plug[2] << tm["worldMatrix"][0]
 
-            >>> plug[2].connection(plug=True) == tm["worldMatrix"]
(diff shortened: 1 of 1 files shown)


monkeez:
The `inputs` method was from the `Node` class which could have multiple inputs.

mottosso:
> The `inputs` method was from the `Node` class which could have multiple inputs.

Oh yes.

mottosso:
These latest tweaks should have gone into its own PR, but sometimes one can be lazy.

- Return native Quaternion types for asQuaternion() and inverse()
- Make Divider attributes non-keyable per default
- Append to existing animation cuves if they exist
- Add minTime(), maxTime(), animationStartTime(), animationEndTime()
- Add Vector.isEquivalent
- Add Quaternion.asEulerRotation
- Add Quaternion.inverse
- Add TransformationMatrix.setRotatePivot
- Print rotation of TransformationMatrix in degrees rather than radians

PR #63, opened 2021-06-20
Fix Node.isA bug when checking multiple types

monkeez:
Because Maya is a big dumb dumb and errors out if you dare compare an MTypeId with anything other than another MTypeId:

```python
node = cmdx.createNode("transform")
node.isA((cmdx.tTransform, cmdx.tMesh))
# Error: MTypeId expected
# Traceback (most recent call last):
#   File "<maya console>", line 2, in <module>
#   File "C:\Users\mitch\code\cmdx\cmdx.py", line 844, in isA
#     return self._fn.typeName in type or self._fn.typeId in type
# TypeError: MTypeId expected # 
```

Let me know if there;s a better way to solve this!

### DIFF
--- a/cmdx.py
+++ b/cmdx.py
@@ -841,7 +841,12 @@ class Node(object):
         elif isinstance(type, string_types):
             return type == self._fn.typeName
         elif isinstance(type, (tuple, list)):
-            return self._fn.typeName in type or self._fn.typeId in type
+            for t in type:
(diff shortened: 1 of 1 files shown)


mottosso:
Nice catch @monkeez!

monkeez:
Ah recursion, I always forget to use it because it confused me when first learning Python.

mottosso:
It's rarely useful.. but sometimes it is!

PR #65, opened 2021-07-05
Walk connections and other tweaks

monkeez:
## walkConnections
Walk through DG connections upstream or downstream starting from either a `Plug` or a `Node`. Defaults to depth first search but can be changed to breadth first.

```python
a = cmdx.createNode("transform")
b = cmdx.createNode("multiplyDivide")
c = cmdx.createNode("transform")
d = cmdx.createNode("transform")
a["tx"] >> b["input1X"]
b["outputX"] >> c["tx"]
a["ty"] >> d["ty"]
list(a.walkConnections())
# Result: [multiplyDivide1, |transform2, |transform3] # 
```

The `type` argument limits the return type.

```python
list(a.walkConnections(type="transform"))
# Result: [|transform2, |transform3] # 
```

Whereas `filter` prunes the iterator, stopping traversal over incorrect node types.

```python
list(a.walkConnections(filter="transform")) # Don't traverse over the multiplyDivide node
# Result: [|transform3] #
```

Return plugs with `plugs`

```python
[x.path() for x in a.walkConnections(plugs=True)]
# Result: ['multiplyDivide1.input1X', '|transform2.translateX', '|transform3.translateY'] #
```

`connections` will return a tuple of both plugs/nodes in the connection.

```python
it = a.walkConnections(connections=True)
next(it)
# Result: (multiplyDivide1, |transform1) # 

it = a.walkConnections(plugs=True, connections=True)
con = next(it)
con[0].path(), con[1].path()
# Result: ('multiplyDivide1.input1X', '|transform1.translateX') # 
```

`level` controls the level of detail when traversing the DG, the default kPlugLevel will return multiple connections between two nodes, whereas kNodeLevel will only return the first connection between two nodes.

`upstream` and `downstream` are also provided as convenience methods for `walkConnections`.

## Plug methods for getting parent plugs
- `isElement` and `isChild` for checking if this plug is part of an array or compound.
- `parent` returns the parent of this element or child plug.
- `index` returns the index of this element or child plug.

## Extra arguments to Node.descendents
- `traversal` for choosing between depth or breadth first.
- `query` for querying attributes like in `Node.children`.
- `contains` for only returning node with a certain shape.

Side note, should it be spelled descendants not descendents?

## Node.query
I've pulled the `query` argument from `Node.children` so it can be used on its own and maintained in the one place. It returns `True` if the query is successful otherwise `False`.

## MatrixType.\_\_invert\_\_
Supports the `~` operator for inverting a matrix.

## Added some missing PEP8 aliases for Plug
There were a couple of method/properties lacking a PEP8 alias.

---

While working on these I noticed a few methods had a `type` argument for checking node types as well as a `filter` argument for checking if the object has a MFn function set. Could these not be rolled into a single `type` argument with a `Node.isA` check? That way you can have exact type checking against a `MTypeId` or `typeName`, or more permissive type checking against an MFn constant. It would also allow type checking against a tuple of types, which isn't possible with some methods at the moment.

I've noticed this `filter` argument in these methods:
- `Node.parent`
- `Node.lineage`
- `Node.children`
- `Node.child`
- `Node.siblings`
- `Node.sibling`

I'd be happy to refactor these methods to use `Node.isA` if you agree and update this PR.

### DIFF
--- a/cmdx.py
+++ b/cmdx.py
@@ -1378,113 +1378,6 @@ class Node(object):
                              destination=True,
                              connections=connection), None)
 
-    def walkConnections(self,
-                        type=None,
-                        filter=None,
-                        direction=ItDg.kDownstream,
-                        traversal=ItDg.kDepthFirst,
-                        level=ItDg.kNodeLevel,
-                        plugs=False,
-                        connections=False):
-        """Walk connections either upstream or downstream from this node
-
-        Arguments:
-            type (MTypeId, str, int, optional): Return only nodes of this type
-            filter (MTypeId, str, int, optional): Iterate over only nodes of this type
-            direction (int, optional): Walk upstream or downstream through the graph
-            traversal (int, optional): Traversal method, depth or breadth first
-            level (int, optional): Level of iteration through the graph,
-                defaults to node level. Iterating at plug level will return
-                every connection between this plug and another node, whereas
-                node level would only return one connection to another node.
-            plugs (bool, optional): Return plugs, rather than nodes
-            connections (bool, optional): Return tuples of the connected plugs
-
-        Example:
-            >>> a = createNode("transform")
-            >>> b = createNode("multiplyDivide")
-            >>> c = createNode("transform")
-            >>> d = createNode("transform")
-            >>> a["tx"] >> b["input1X"]
-            >>> b["outputX"] >> c["tx"]
-            >>> a["ty"] >> d["ty"]
-            >>> cons = list(a.walkConnections())
-            >>> all(x in cons for x in (b, c, d))
-            True
-            >>> it = a.walkConnections(filter="transform")
-            >>> next(it) == d
-            True
-            >>> it = c.walkConnections(type="transform", direction=ItDg.kUpstream)
-            >>> next(it) == a
-            True
-            >>> cons = list(a.walkConnections(plugs=True, connections=True))
-            >>> mplugs = [(x.plug(), y.plug()) for x, y in cons]
-            >>> (b["input1X"].plug(), a["tx"].plug()) in mplugs
-            True
-            >>> (c["tx"].plug(), b["outputX"].plug()) in mplugs
-            True
-            >>> (d["ty"].plug(), a["ty"].plug()) in mplugs
-            True
-
-        """
-
-        it = ItDg(self._mobject, direction=direction, traversal=traversal, level=level)
-
-        while not it.isDone():
-            currentNode = it.currentNode()
-
-            if currentNode == self._mobject: # Skip self
-                it.next()
-                continue
-
-            node = Node(currentNode)
-
-            if filter is not None and not node.isA(filter):
-                it.prune()
-
-            elif type is None or node.isA(type):
-                connection = Plug(node, it.currentPlug()) if plugs else node
-                if connections:
-                    plug = it.previousPlug()
-                    previous = Node(plug.node())
-                    yield connection, Plug(previous, plug) if plugs else previous
-                else:
-                    yield connection
-
-            it.next()
-
-    def downstream(self,
-                   type=None,
-                   filter=None,
-                   traversal=ItDg.kDepthFirst,
-                   level=ItDg.kPlugLevel,
-                   plugs=False,
-                   connections=False):
-        """Walk connections downstream from :func:`walkConnections()`"""
-        return self.walkConnections(type=type,
-                                    filter=filter,
-                                    direction=ItDg.kDownstream,
-                                    traversal=traversal,
-                                    level=level,
-                                    plugs=plugs,
-                                    connections=connections)
-
-    def upstream(self,
-                 type=None,
-                 filter=None,
-                 traversal=ItDg.kDepthFirst,
-                 level=ItDg.kPlugLevel,
-                 plugs=False,
-                 connections=False):
-        """Walk connections upstream from :func:`walkConnections()`"""
-        return self.walkConnections(type=type,
-                                    filter=filter,
-                                    direction=ItDg.kUpstream,
-                                    traversal=traversal,
-                                    level=level,
-                                    plugs=plugs,
-                                    connections=connections)
-
     def rename(self, name):
         mod = om.MDGModifier()
         mod.renameNode(self._mobject, name)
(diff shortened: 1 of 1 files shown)


mottosso:
Nice one! Some questions.

> list(a.walkConnections())

What does walking connections starting from a *node* mean? Would it not make more sense starting from an attribute?

```py
a["tx"].walk()
```

> The type argument limits the return type ... Whereas filter prunes the iterator, stopping traversal over incorrect node types.

Hm, it makes sense to offer both of these abilities, but that type and filter would act differently here is not obvious. Is there another term we could use? E.g. `stopCondition` or `breakPoint`?

> `level` controls the level of detail when traversing the DG

How about the keyword `depth` instead? I think it does a better job communicating the number of traversals being made.

> Side note, should it be spelled descendants not descendents?

Yes you are right! Let's rename it, but leave an alias below `descendents = descendants` for folk like me who can't spell.

> Could these not be rolled into a single type argument with a Node.isA check?

Not sure, try it.

mottosso:
> How about the keyword depth instead?

Come to think of it, `maxDepth` really hits home I think. No ambiguity there, as `depth` could mean that you're looking for a connection at a specific depth, e.g. `2` for the 2nd connection.

monkeez:
> What does walking connections starting from a node mean? Would it not make more sense starting from an attribute?

I've implemented `Plug.walkConnections` and `Node.walkConnections` so both are available. But to answer your question `Node.walkConnections` works in a similar way to `Node.connections` in that it will iterate over its plugs that have connections, the difference being from there it can walk further through the DG.

`filter` and `level` were taken from the argument names for MItDependencyGraph. Happy to change them if we find something more clear.

I'm not sure `stopCondition` or `breakPoint` quite work in place of `filter`, they make me think that the iterator will stop if some condition is true. Whereas the argument only lets the iterator continue down a limb of the tree if a condition is true, otherwise it moves on to the next limb if there are any. I'll have a think on other options.

`level` isn't a depth check, it changes the level of detail of the iteration (node level or plug level). Given the following connections:

```
  nodeA            nodeB
 ________         ________
|        |       |        |
|     tx o ----- o tx     |
|     ty o       o ty     |
|     tz o ----- o tz     |
|________|       |________|
```

If we walk connections from `nodeA` in node level of detail:

```python
list(nodeA.walkConnections(level=cmdx.ItDg.kNodeLevel))
# Result: [nodeB] #
```

Whereas walking at plug level returns both connections to nodeB:

```python
list(nodeA.walkConnections(level=cmdx.ItDg.kPlugLevel))
# Result: [nodeB, nodeB] #
```

Node.walkConnections defaults to node level, whereas Plug.walkconnections defaults to plug level.

monkeez:
After using `walkConnections` in some other code I feel like there's some improvements I can make for usability. Hopefully be able to update this PR shortly.

mottosso:
Thanks for elaborating. I noticed these were terminology borrowed from the API, and it does make sense to stick with it. Even though I really don't find it intuitive. :S

My main concerns are in `type` and `filter` sounding very similar, but acting differently. If there's a way around that, that would be great. `level` tells me nothing, but if it's from the API then at least there's partial familiarity with anyone who's dabbled with that. It's not depth-related as I thought, so that's good. Would be good to have a `maxDepth` however, I often find myself traversing connections but aren't interested in anything 3-4 iterations away. That said, since it's an iterator the user can always `break` without paying the full cost of traversing the entire thing. That's probably the more Pythonic way to go about it.

> After using walkConnections in some other code I feel like there's some improvements I can make for usability. Hopefully be able to update this PR shortly.

Cool. Is your code public? I would be curious to see how other people's code actually looks like. One of the things that I've had to painstakingly change was getting rid of all `node["attr"] = value` in favour of `mod.set_attr(node["attr"], value)` to support undo. Meaning I never actually benefit from that nice syntax, nor for connecting with `node["a"] >> node["b"]`.

monkeez:
> Cool. Is your code public?

Nothing public atm, but would like to release some stuff. I'm habitually starting things and not finishing them as coding is mostly a hobby for me at this point.

Did `node["attr"] = value` used to be undoable in a previous release?

> My main concerns are in type and filter sounding very similar, but acting differently.

A possibility is removing the `type` argument and leaving it up to the user to only use the types they're after.

mottosso:
> Did node["attr"] = value used to be undoable in a previous release?

No no, it never was. Undo is almost entirely based on `cmdx.DagModifier`. It just isn't as syntactically nice to look at. Something like [this](https://github.com/mottosso/ragdoll/blob/4ed4e7f831b81fe68e643cd6d342c06d68515b27/ragdoll/commands.py#L56) for example. Lots of repetitive `mod.something()` calls. :((

> coding is mostly a hobby for me at this point.

Stellar code quality for a hobbyist, I must say!

monkeez:
Hey so I've updated this PR.

The issue I was having with the way I had originally written `walkConnections` was that it was fairly inflexible. If you wanted to do more complex checks than a simple type check, it quickly became not worth using. So instead I've refactored it into `WalkDG`, this works closer to how `MItDependencyGraph` works, except it returns `cmdx` classes, and type checking uses `Node.isA` which is a bit more flexible. I also replaced the `filter` argument with the `pruneOnType` argument, if this is `True` the iterator will prune the tree when it encounters a node that isn't of `type`.

Let me know your thoughts.

mottosso:
Sorry for the delay, back from holiday now and hope to get this integrated by end of the week. 🤞

mottosso:
End of the week is here, and I'm having a closer look at this now. :) The delay has been due to (1) it's a lot of code to review, and (2) it's not something I current use nor need. So there is a lacking incentive to make time for it, sorry about that!

Looking at it now, I'm missing some motivation for this to exist. What can we do now, that we couldn't do before? E.g. there already is `node["attr"].connections()` and `node["attr"].output()` etc. to traverse from one plug to another. I'm missing why we need this PR?

And if we do, why shouldn't we inherit from `MItDependencyGraph` and wrap any output with cmdx instances? It seems we need to invent terminology that already exist with `MItDependencyGraph`.

I'm thinking you've been using this for a while now, what are your usecases? What would you be unable to do without this PR? The PEP8 aliases and `~` for matrices seem solid, I'd like to merge those. But separately. It's mostly `WalkDG` that I'm not seeing a purpose for.

monkeez:
Hey

To be honest I can't remember why I didn't inherit from `MItDependencyGraph`, there was some reason but it's escaping me 😄.

Yeah I haven't ended up finding this to be as useful as I initially thought.

Happy to close this and just make a new PR with the stuff you do want.

mottosso:
Ah ok, yes let's do that then. 👍

PR #70, opened 2021-11-16
include plug indices in their path() and name()

Muream:
Like the title said, I've included the index of the array plugs in their names and path.

Some error messages are using `Plug.path()` which was misleading before this PR because it would only mention the name of the "parent" of the actual plug. (eg. worldMatrix instead of worldMatrix[0])
The ones I ran into are now displaying the correct path.

### DIFF
--- a/cmdx.py
+++ b/cmdx.py
@@ -3614,6 +3614,10 @@ class Plug(object):
                 '|persp.translateX'
                 >>> persp["translateX"].path(full=True)
                 '|persp.translate.translateX'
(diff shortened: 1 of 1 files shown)


mottosso:
Great idea. :) Once the tests pass, this is happening.

Muream:
Sweet 👌 
Is the pypi package updated automatically when something's merged in master?

mottosso:
It's generated once a new tag/release is made, which I'll do now. :)

PR #73, opened 2022-02-28
Clone

mottosso:
Support for cloning enum attributes.

```py
parent = createNode("transform")
camera = createNode("camera", parent=parent)

# Make new enum attribute
camera["myEnum"] = Enum(fields=["a", "b", "c"])

# Clone it
clone = camera["myEnum"].clone("cloneEnum")
cam.addAttr(clone)

# Compare it
fields = camera["cloneEnum"].fields()
assert fields == ((0, "a"), (1, "b"), (2, "c"))
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,8 +22,6 @@ jobs:
 
       matrix:
        include:
-         - maya: "2017"
-           pip: "2.7/get-pip.py"
          - maya: "2018"
            pip: "2.7/get-pip.py"
          - maya: "2019"
@@ -31,6 +29,10 @@ jobs:
          - maya: "2020"
            pip: "2.7/get-pip.py"
          - maya: "2022"
+           pip: "get-pip.py"
+         - maya: "2023"
+           pip: "get-pip.py"
+         - maya: "2024"
            pip: "get-pip.py"
 
     container: mottosso/maya:${{ matrix.maya }}
--- a/README.md
+++ b/README.md
@@ -528,6 +528,50 @@ The reason for this limitation is because the functions `cmds`
 
 <br>
 
+### Path-like Syntax
+
+Neatly traverse a hierarchy with the `|` syntax.
+
+```py
+# Before
+group = cmdx.encode("|some_grp")
+hand = cmdx.encode(group.path() + "|hand_ctl")
+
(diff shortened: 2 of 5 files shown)


mottosso:
A little overdue, some additions that's been happening over the past month or so.

- Support for path-like syntax, e.g. parent | "child" | "grandchild"
- Add Vector.rotate_by PEP8 syntax
- Protect against crash when using MPoint with MObject
- Support BoundingBox.expand with Vector-type
- Expose math.sqrt as cmdx.sqrt
- Expose isScrubbing
- Expose isPlaying
- Support for cmdx.setAttr(undoable=True)
- Support for cmdx.setAttr with Matrix
- Make Maya aware of cmdx version

<br>

### Path-like Syntax

You can now more neatly identify objects in a hierarchy.

```py
# Before
group = cmdx.encode("|some_grp")
hand = cmdx.encode(group.path() + "|hand_ctl")

# After
hand = group | "hand_ctl"
```

It can be nested too.

```py
finger = group | "hand_ctl" | "finger_ctl"
```

<br>

### setAttr

To my great astonishment, Maya's `cmds.setAttr` depends on the UI settings for units. :O

```py
cmds.setAttr("hand_ctl.translateY", 5)
```

For a user with Maya set to `Centimeters`, this would set `translateY` to 5 centimeters. For any user with any other unit, like `Foot`, it would instead move it 5 feet. That is terrible behaviour for a script, how can you possibly define the length of something if you don't know the unit? A dog is 100 cm tall, not 100 "any unit" tall.

The `cmdx.setAttr` on the other hand does what Maya's API does, which is to treat all units consistently.

- Distance values are in centimeters
- Angular values are in radians

So the user is free to choose any unit for their UI without breaking their scripts.

mottosso:
More additions, should really be merging this hah.

- setChannelBox
- currentFrame
- Handle passing tuple to curve()
- Integer attribute type
- Handle multiple vendored cmdx's

mottosso:
We've got company. 😃 https://github.com/AnimalLogic/AL_omx

Bringing this up to date with [the version bundled with Ragdoll](https://github.com/mottosso/ragdoll/blob/main/ragdoll/vendor/cmdx_ragdoll.py).

PR #78, opened 2024-02-17
Implements a different approach for __repr__

chelloiaco:
## Description of Changes
Expanding on #20 , this changes __repr__ to give a clear representation of what the object is, preventing the user from assuming it is a string. The goal, while not as forgiving, is to avoid unpredictable behavior by making the user having to explicitly request the name of a node via `.name()`, its path via `.path()`, and the value of one of its attributes via `.read()`.

**Note**: I opted to have `__str__` be different from `__repr__`, that way we still have the ability to `str(node)` and get the previous behavior, as seen by the changes done on **[tests.py](https://github.com/mottosso/cmdx/compare/master...chelloiaco:cmdx:repr-changes#diff-f455f302936271d755f6892d443786094d2607b505ff180ca817cb9eeabe1e5eL384-R388)**. I'm not certain if this would be confusing or unpredictable, but let me know.

## Testing done
As expected, I had to change the doctests to utilize `.name()`, `.path()` or `.read()` accordingly. After the changes, those pass without failures:
```
===Flaky Test Report===

test_createNode_performance passed 1 out of the required 3 times. Running test again until it passes 3 times.
test_createNode_performance passed 2 out of the required 3 times. Running test again until it passes 3 times.
test_createNode_performance passed 3 out of the required 3 times. Success!
test_rouge_mode passed 1 out of the required 3 times. Running test again until it passes 3 times.
test_rouge_mode passed 2 out of the required 3 times. Running test again until it passes 3 times.
test_rouge_mode passed 3 out of the required 3 times. Success!

===End Flaky Test Report===
Ran 153 tests in 1.861s

OK
Skipping coveralls
```

### DIFF
--- a/cmdx.py
+++ b/cmdx.py
@@ -2674,20 +2674,27 @@ class Plug(object):
         return str(self.read())
 
     def __repr__(self):
-        cls_name = '{}.{}'.format(__name__, self.__class__.__name__)
-        if self._mplug.attribute().apiType() == om.MFn.kCompoundAttribute:
-            return '{}("{}", "{}")'.format(cls_name,
-                                           self.node().name(),
-                                           self.name())
-        read_val = self.read()
-        if isinstance(read_val, string_types):
-            # Add surrounding single quotes, indicating the value is a string
-            read_val = '"{}"'.format(read_val)
+        try:
+            # Delegate the value reading to __str__
+            read_result = str(self)
+            valid = True
+        except:
+            valid = False
 
-        return '{}("{}", "{}") == {}'.format(cls_name,
-                                             self.node().name(),
-                                             self.name(),
-                                             read_val)
+        cls_name = '{}.{}'.format(__name__, self.__class__.__name__)
+        msg = '{}("{}", "{}")'.format(cls_name,
+                                      self.node().name(),
+                                      self.name())
+        if valid:
+            try:
+                if self.typeClass() == String:
+                    # Add surrounding quotes, indicating it is a string
+                    read_result = '"{}"'.format(read_result)
+            except TypeError:
+                pass
+            msg += ' == {}'.format(read_result)
+
+        return msg
 
     def __rshift__(self, other):
         """Support connecting attributes via A >> B"""
--- a/cmdx.py
+++ b/cmdx.py
@@ -511,6 +511,8 @@ class Node(object):
         >>> transform = createNode("transform")
         >>> transform["tx"] = 5
         >>> transform["worldMatrix"][0] >> decompose["inputMatrix"]
+        >>> decompose["outputTranslate"]
+        <cmdx.Plug : (5.0, 0.0, 0.0)>
         >>> decompose["outputTranslate"].read()
         (5.0, 0.0, 0.0)
 
@@ -579,6 +581,8 @@ class Node(object):
         Example:
             >>> node = createNode("transform")
             >>> node["translate"] = (1, 1, 1)
+            >>> node["translate", Meters]
+            <cmdx.Plug : (0.01, 0.01, 0.01)>
             >>> node["translate", Meters].read()
             (0.01, 0.01, 0.01)
 
@@ -625,6 +629,8 @@ class Node(object):
             True
             >>> node["rotateX", Degrees] = 1.0
             >>> node["rotateX"] = Degrees(1)
+            >>> node["rotateX", Degrees]
+            <cmdx.Plug : 1.0>
             >>> node["rotateX", Degrees].read()
             1.0
             >>> node["myDist"] = Distance()
@@ -981,6 +987,8 @@ class Node(object):
         Examples:
             >>> node = createNode("transform")
             >>> node.update({"tx": 5.0, ("ry", Degrees): 30.0})
+            >>> node["tx"]
+            <cmdx.Plug : 5.0>
             >>> node["tx"].read()
             5.0
 
@@ -999,13 +1007,19 @@ class Node(object):
         Example:
             >>> node = createNode("transform")
             >>> node["translateX"] = 5
+            >>> node["translateX"]
+            <cmdx.Plug : 5.0>
             >>> node["translateX"].read()
             5.0
             >>> # Plug was reused
+            >>> node["translateX"]
+            <cmdx.Plug : 5.0>
             >>> node["translateX"].read()
             5.0
             >>> # Value was reused
             >>> node.clear()
+            >>> node["translateX"]
+            <cmdx.Plug : 5.0>
             >>> node["translateX"].read()
             5.0
             >>> # Plug and value was recomputed
@@ -1524,8 +1538,9 @@ class DagNode(Node):
             >>> _new()
             >>> parent = createNode("transform", "parent")
             >>> child =  createNode("transform", "child", parent)
-            >>> result = (parent | "child").path()
-            >>> result in (
+            >>> parent | "child"
+            <cmdx.DagNode : '|parent|child'>
+            >>> (parent | "child").path() in (
             ...    '|parent|child',
             ...   u'|parent|child'
             ... )
@@ -1533,8 +1548,9 @@ class DagNode(Node):
 
             # Stackable too
             >>> grand =  createNode("transform", "grand", child)
-            >>> result = (parent | "child" | "grand").path()
-            >>> result in (
+            >>> parent | "child" | "grand"
+            <cmdx.DagNode : '|parent|child|grand'>
+            >>> (parent | "child" | "grand").path() in (
             ...    '|parent|child|grand',
             ...   u'|parent|child|grand'
             ... )
@@ -2301,22 +2317,24 @@ class ObjectSet(Node):
         """Return members, converting nested object sets into its members
 
         Example:
-            >>> from maya import cmds
-            >>> _ = cmds.file(new=True, force=True)
-            >>> a = cmds.createNode("transform", name="a")
-            >>> b = cmds.createNode("transform", name="b")
-            >>> c = cmds.createNode("transform", name="c")
-            >>> cmds.select(a)
-            >>> gc = cmds.sets([a], name="grandchild")
-            >>> cc = cmds.sets([gc, b], name="child")
-            >>> parent = cmds.sets([cc, c], name="parent")
-            >>> mainset = encode(parent)
-            >>> result = sorted([n.path() for n in mainset.flatten()])
-            >>> result in (
-            ...    ['|a', '|b', '|c'],
-            ...    [u'|a', u'|b', u'|c']
-            ... )
-            True
+          >>> from maya import cmds
+          >>> _ = cmds.file(new=True, force=True)
+          >>> a = cmds.createNode("transform", name="a")
+          >>> b = cmds.createNode("transform", name="b")
+          >>> c = cmds.createNode("transform", name="c")
+          >>> cmds.select(a)
+          >>> gc = cmds.sets([a], name="grandchild")
+          >>> cc = cmds.sets([gc, b], name="child")
+          >>> parent = cmds.sets([cc, c], name="parent")
+          >>> mainset = encode(parent)
+          >>> sorted(mainset.flatten(), key=lambda n: n.name())
+          [<cmdx.DagNode : '|a'>, <cmdx.DagNode : '|b'>, <cmdx.DagNode : '|c'>]
+          >>> result = sorted([n.name() for n in mainset.flatten()])
+          >>> result in (
+          ...    ['a', 'b', 'c'],
+          ...    [u'a', u'b', u'c']
+          ... )
+          True
 
         """
 
@@ -2574,8 +2592,12 @@ class Plug(object):
         Example:
             >>> node = createNode("transform")
             >>> node["tx"] = 5
+            >>> node["translate"] + "X"
+            <cmdx.Plug : 5.0>
             >>> (node["translate"] + "X").read()
             5.0
+            >>> node["t"] + "x"
+            <cmdx.Plug : 5.0>
             >>> (node["t"] + "x").read()
             5.0
             >>> try:
@@ -2611,10 +2633,16 @@ class Plug(object):
             >>> node["myArray"].extend([2.0, 3.0])
             >>> node["myArray"] += 5.1
             >>> node["myArray"] += [1.1, 2.3, 999.0]
+            >>> node["myArray"][0]
+            <cmdx.Plug : 1.0>
             >>> node["myArray"][0].read()
             1.0
+            >>> node["myArray"][6]
+            <cmdx.Plug : 999.0>
             >>> node["myArray"][6].read()
             999.0
+            >>> node["myArray"][-1]
+            <cmdx.Plug : 999.0>
             >>> node["myArray"][-1].read()
             999.0
 
@@ -2818,6 +2846,8 @@ class Plug(object):
         Example:
             >>> node = createNode("transform")
             >>> node["translate"][0] = 5
+            >>> node["tx"]
+            <cmdx.Plug : 5.0>
             >>> node["tx"].read()
             5.0
 
@@ -3193,8 +3223,12 @@ class Plug(object):
             >>> node = createNode("transform")
             >>> node["myArray"] = Double(array=True)
             >>> node["myArray"].extend([1.0, 2.0, 3.0])
+            >>> node["myArray"][0]
+            <cmdx.Plug : 1.0>
             >>> node["myArray"][0].read()
             1.0
+            >>> node["myArray"][-1]
+            <cmdx.Plug : 3.0>
             >>> node["myArray"][-1].read()
             3.0
 
@@ -4121,6 +4155,8 @@ class Plug(object):
             True
             >>> b["ihi"].connection() == a
             True
+            >>> a["ihi"]
+            <cmdx.Plug : 2>
             >>> a["ihi"].read()
             2
             >>> b["arrayAttr"] = Long(array=True)
@@ -6370,8 +6406,9 @@ class _BaseModifier(object):
             >>> with DagModifier() as mod:
             ...     mod.connect(tm["sx"], tm["tx"])
             ...
-            >>> result = tm["tx"].connection().path()
-            >>> result in (
+            >>> tm["tx"].connection()
+            <cmdx.DagNode : '|myTransform'>
+            >>> tm["tx"].connection().path() in (
             ...    '|myTransform',
             ...   u'|myTransform'
             ... )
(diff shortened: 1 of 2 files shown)


cmdx.py
```diff
@@ -1522,13 +1524,13 @@ def __or__(self, other):
...
-            |parent|child
+            >>> (parent | "child").path()
+            '|parent|child'
 
             # Stackable too
             >>> grand =  createNode("transform", "grand", child)
-            >>> parent | "child" | "grand"
-            |parent|child|grand
+            >>> (parent | "child" | "grand").path()
+            '|parent|child|grand'
```
chelloiaco, line 1533: BY the way, it might be a bit hacky, but we could make this more cmdx friendly by adding something like the following to `__or__`:
```python
if not isinstance(other, string_types):
    if isinstance(other, DagNode):
        other = other.name()
```
Then, we could provide DagNodes and this would work:
```
>>> parent | child | grand
<cmdx.DagNode : '|parent|child|grand'>

>>> (parent | child | grand).path()
'|parent|child|grand'
```

cmdx.py
```diff
@@ -511,7 +511,7 @@ class Node(object):
         >>> transform = createNode("transform")
         >>> transform["tx"] = 5
         >>> transform["worldMatrix"][0] >> decompose["inputMatrix"]
-        >>> decompose["outputTranslate"]
+        >>> decompose["outputTranslate"].read()
```
mottosso, line 514: Hmm, having to put `.read()` here is a bummer. I can't think of any scenario where this affects code outside of these doctests, but at least we should include a few with the result without `.read()` so we can see what they look like. Can you include a few?

chelloiaco: Another option would be to revert these and just change the output to accommodate the new repr result. Did you still want to have both?

cmdx.py
mottosso: Huh? Wouldn't `grand` then already be a `cmdx.DagNode` at which point.. there would be no need to use `|`? 🤔 Maybe if you put together a complete example would it be easier to see what you've got in mind.

cmdx.py
mottosso: A mixture would be best, we both want to test these outputs and also document them.

cmdx.py
chelloiaco: > Wouldn't grand then already be a cmdx.DagNode at which point.. there would be no need to use |? 

You could still use the string version of it, in combination with DagNodes:
```python
>>> parent = cmdx.createNode("transform", "parent")
>>> child =  cmdx.createNode("transform", "child", parent)
>>> grand =  cmdx.createNode("transform", "grand", child)
>>> parent | "child" | grand
<cmdx.DagNode : '|parent|child|grand'>
```

Although, having a second thought on this, it is probably unnecessary. I thought it could be nice to have some flexibility, but it perhaps would just add confusion.

I guess the power of using the strings here is to be able to do things like:
```python
>>> parent | "chi*"
<cmdx.DagNode : '|parent|child'>
>>> parent | "chi*" | "*"   
<cmdx.DagNode : '|parent|child|grand'>
```

cmdx.py
```diff
@@ -2291,18 +2317,24 @@ def flatten(self, type=None):
...
+          >>> a = cmds.createNode("transform", name="a")
+          >>> b = cmds.createNode("transform", name="b")
+          >>> c = cmds.createNode("transform", name="c")
+          >>> cmds.select(a)
+          >>> gc = cmds.sets([a], name="grandchild")
+          >>> cc = cmds.sets([gc, b], name="child")
+          >>> parent = cmds.sets([cc, c], name="parent")
+          >>> mainset = encode(parent)
+          >>> sorted(mainset.flatten(), key=lambda n: n.name())
+          [<cmdx.DagNode : '|a'>, <cmdx.DagNode : '|b'>, <cmdx.DagNode : '|c'>]
```
chelloiaco, line 2331: Un-indented these by 2 spaces so line 2331 isn't too long.

mottosso: There we go, now I can actually see what these look like. What do you think about:

```py
# Before
<cmdx.DagNode : '|node'>

# After
<cmdx.DagNode("|node")>
```

That way, it'll be more familiar and copy-pasteable as real code too.

chelloiaco:
> There we go, now I can actually see what these look like. What do you think about:
> 
> ```python
> # Before
> <cmdx.DagNode : '|node'>
> 
> # After
> <cmdx.DagNode("|node")>
> ```
> 
> That way, it'll be more familiar and copy-pasteable as real code too.

I did think at one point to use parenthesis similarly to your suggestion, but decided not to go with it exactly ***because*** it started to look like executable code.

The issue is that there isn't a matching initializer for DagNode (or Node) that takes in a string name. I'm thinking this could potentially be confusing? However, it would be easy enough to implement if we wanted to go that route.

That said, it gets a bit more complicated for Plug nodes if we want copy-pasteable code. Considering node name initialization is supported, it wouldn't be helpful to have:
```console
>>> node["translate"]
<cmdx.Plug("|node", "translate") = (0.0, 0.0, 0.0)>
```

This is because assignment operator (`=`) to a function call is syntactically invalid in Python and therefore the provided code snippet wouldn't be executable. A valid solution would be:
```console
>>> node["translate"]
<cmdx.Plug("|node", "translate").write((0.0, 0.0, 0.0))>
```
But, while it shows what could be executable code, this starts becoming cryptic and isn't a concise representation of the object.

Let me know your thoughts. What I used was really just an initial suggestion to get the idea across, by no means it's the best solution!

mottosso:
Hmm, a lot of minutia here.

I had a look at how PyMEL does it; I've never noticed it before, which is a good sign.

```py
from pymel import core as pm
t = pm.createNode("transform", name="myTransform")
print(repr(t))
# nt.Transform('myTransform')
```

It looks like an unrelated instance, due to the `nt` namespace, which is unexpected. It also does not include hierarchy in the name, which I prefer. I would expect that if a user wanted to know the path, rather than name, they would not print the node but ask for its `.path()`. Displaying the type and name when printing sounds more intuitive to me.

For plugs..

```py
print(repr(t.translateX))
# Attribute('myTransform.translateX')
```

Now there is no namespace at all, which is also unexpected. There is also no value, which is unexpected also; why print a value if it isn't going to give you the value?

But I do like the presence of the object and attribute name, so let's keep that.

How about.

```py
>>> node
cmdx.DagNode("myTransform")
>>> node
cmdx.Node("multMatrix1")
>>> node["translateX"]
cmdx.Plug("myTransform", "translateX") ==  0.0
>>> node["translate"]
cmdx.Plug("myTransform", "translate") == (0.0, 0.0, 0.0)
```

Where we:

1. Use double-quoted
2. Skip the `<>` prefix/suffix
3. Skip the hierarchy of the dag node
4. Use `==` to denote the value of a plug

Later on, we can discuss whether or not to make this actually executable; for now, it reads well I think.

chelloiaco:
> ```python
> >>> node
> cmdx.DagNode("myTransform")
> >>> node
> cmdx.Node("multMatrix1")
> >>> node["translateX"]
> cmdx.Plug("myTransform", "translateX") ==  0.0
> >>> node["translate"]
> cmdx.Plug("myTransform", "translate") == (0.0, 0.0, 0.0)
> ```

I've changed the reprs to display this way and changed the tests accordingly. All seems to be working good and passing the tests.

There is one thing that I ran into though, which is due to how .read() works:
```python
>>> _ = cmds.file(new=1, f=1)
>>> sphere_a = cmds.polySphere(ch=False)[0]
>>> sphere_b = cmds.polySphere(ch=False)[0]
>>> bs = cmds.blendShape(sphere_a, sphere_b)[0]
>>> bs = cmdx.ls(bs)[0]
>>> bs
cmdx.Node("blendShape1")
>>> bs['input']
cmdx.Plug("blendShape1", "input") == (None, 0, '*')
```

What's happening here is that its `read()`ing the values for `'input[0].inputGeometry'`, `input[0].groupId'` and `'input[0].componentTagExpression'`. Might be an edge-case but looks kinda weird. A solution would be to add checks to only read() if the type is worth reading. But, then we would probably be over-complicating the __repr__

<br>

_PS: I feel like I might've opened a can of worms, but running `bs['inputTarget']` causes an API failure when it tries to get 'targetBindMatrix'. I might create a separate issue on it so we keep on topic here._

mottosso:
> cmdx.Plug("blendShape1", "input") == (None, 0, '*')

You mean because `input` is a compound attribute? Maybe we could limit value printing to simple types - float, double, bool and string?

Other than that, I think it looks good!

cmdx.py
```diff
@@ -2674,12 +2674,16 @@ def __str__(self):
         return str(self.read())
 
     def __repr__(self):
+        cls_name = '{}.{}'.format(__name__, self.__class__.__name__)
+        if self._mplug.attribute().apiType() == om.MFn.kCompoundAttribute:
```
mottosso, line 2678: I would go the opposite route; include rather than exclude. There are lots of types, and if we aren't going to test them all then we should only include the ones we know work. I expect printing meshes and MObject types to also be problematic. Only the float, double, bool and string are interesting, possible enum too.

cmdx.py
```diff
@@ -2627,7 +2674,20 @@ def __str__(self):
...
+                                           self.name())
+        read_val = self.read()
+        if isinstance(read_val, string_types):
+            # Add surrounding single quotes, indicating the value is a string
+            read_val = '"{}"'.format(read_val)
+
+        return '{}("{}", "{}") == {}'.format(cls_name,
+                                             self.node().name(),
+                                             self.name(),
+                                             read_val)
```
chelloiaco, line 2690: Alright, added the check here to not include the `read()`ing of the values in case the attribute is compound.

The previous example now looks like the following:
```python
>>> _ = cmds.file(new=True, f=True)
>>> sphere_a = cmds.polySphere(ch=False)[0]
>>> sphere_b = cmds.polySphere(ch=False)[0]
>>> bs = cmds.blendShape(sphere_a, sphere_b)[0]
>>> bs = cmdx.ls(bs)[0]

>>> bs
cmdx.Node("blendShape1")

>>> bs['input']  # A kCompoundAttribute, won't read/show values
cmdx.Plug("blendShape1", "input")

>>> sphere_b = cmdx.ls(sphere_b)[0] 
>>> sphere_b['translate']
cmdx.Plug("pSphere2", "translate") == (0.0, 0.0, 0.0)
```

cmdx.py
chelloiaco: Instead of adding more checks into the repr, I thought, since `self.typeClass()` is doing all the heavy lifting, I should use that info instead.

I had to add a couple more types since a lot of these are types of some basic queries like `node['transform']` or `node['rotateX']` but this is what I came up with:
```python
...
valid_types = (Angle, Boolean, Distance, Double, Double3, Enum, Float,
               Integer, Long, Matrix, String)
if self.typeClass() not in valid_types:
    return '{}("{}", "{}")'.format(cls_name,
                                    self.node().name(),
                                    self.name())
...
```

Which worked, with the exception of when the type has not been implemented, since `self.typeClass()` raises an error. This happens specifically with the following doctest on line 4180:
```python
>>> _ = cmds.file(new=True, force=True)
>>> a = createNode("transform", name="A")
>>> b = createNode("multDoubleLinear", name="B")
>>> a["ihi"] << b["ihi"]
>>> a["ihi"].connection() == b
True
>>> b["ihi"].connection() == a
True
>>> a["ihi"]  # Error here
```

That is because `isHistoricallyInteresting` is a kNumericAttribute of `kByte` NumericData, which isn't supported in `self.typeClass()`

I could add a try except  in the __repr__ like so and then not read() the value:
```python
...
typ_class = None
try:
    typ_class = self.typeClass()
except TypeError:
    pass

if typ_class not in valid_types:
...
```

But then the doctest above fails for a different reason:
```python
# expected
cmdx.Plug("A", "isHistoricallyInteresting") == 2

# actual repr return
cmdx.Plug("A", "isHistoricallyInteresting")
```

Would it be ok to remove the value return from that test? Also, in regards to having a try except in the `__repr__`, I'm not so sure about that idea either...

mottosso: Hm, true.. Choices choices.

In that case, I think we can just put whatever comes out of `str(plug)` after the `==`. It can be a try/except; if it fails, we do not include the `==`. It's a `__repr__` after all, for debugging. Not performance sensitive or in need to be exact. Then we could potentially work, separately, on making more values print nicely via `str(plug)`

cmdx.py
```diff
@@ -2674,20 +2674,27 @@ def __str__(self):
...
+                                      self.node().name(),
+                                      self.name())
+        if valid:
+            try:
+                if self.typeClass() == String:
+                    # Add surrounding quotes, indicating it is a string
+                    read_result = '"{}"'.format(read_result)
+            except TypeError:
+                pass
+            msg += ' == {}'.format(read_result)
```
chelloiaco, line 2695: Had to add this in order to avoid cases where it shows a valid value, but of an empty string:
```python
>>> myNode["myStr"]
cmdx.Plug("myNode", "myStr") == 
```

chelloiaco:
> There is one thing that I ran into though, which is due to how .read() works:
> 
> ```python
> >>> _ = cmds.file(new=1, f=1)
> >>> sphere_a = cmds.polySphere(ch=False)[0]
> >>> sphere_b = cmds.polySphere(ch=False)[0]
> >>> bs = cmds.blendShape(sphere_a, sphere_b)[0]
> >>> bs = cmdx.ls(bs)[0]
> >>> bs
> cmdx.Node("blendShape1")
> >>> bs['input']
> cmdx.Plug("blendShape1", "input") == (None, 0, '*')
> ```
This new commit brings back the behavior above for compound attributes (if they don't happen to raise an error). But, I do like the idea of having `__str__` (or perhaps `self.read`?) deal with it. Two (three?) birds one stone.

cmdx.py
```diff
@@ -2627,7 +2674,27 @@ def __str__(self):
         return str(self.read())
 
     def __repr__(self):
-        return str(self.read())
+        try:
+            # Delegate the value reading to __str__
+            read_result = str(self)
```
mottosso, line 2679: Hmmmm, now str() is a call I would never expect to fail, and if it does then I'd consider that a bug. We can expect it will never fail, and not bother with a `valid = True`.

cmdx.py
```diff
@@ -2627,7 +2674,27 @@ def __str__(self):
...
+        except:
+            valid = False
+
+        cls_name = '{}.{}'.format(__name__, self.__class__.__name__)
+        msg = '{}("{}", "{}")'.format(cls_name,
+                                      self.node().name(),
+                                      self.name())
+        if valid:
+            try:
+                if self.typeClass() == String:
```
mottosso, line 2690: Since we've established that typeClass isn't relevant for all types, like kByte, how about we use Maya's apiType here instead? Especially since we're only dealing with a single type.

- https://github.com/mottosso/cmdx/blob/344aed050aa8dec79566d479de354f4a94ff0d4b/cmdx.py#L3868

It would save on the use of try/except too.

Did we also have a test for this case?

cmdx.py
chelloiaco: > PS: I feel like I might've opened a can of worms, but running bs['inputTarget'] causes an API failure when it tries to get 'targetBindMatrix'. I might create a separate issue on it so we keep on topic here.

This is the reason why I added a try/except. running the following causes an API failure:

```python
>>> _ = cmds.file(new=1, f=1)
>>> sphere_a = cmds.polySphere(ch=False)[0]
>>> sphere_b = cmds.polySphere(ch=False)[0]
>>> bs = cmds.blendShape(sphere_a, sphere_b)[0]
>>> bs = cmdx.ls(bs)[0]
>>> bs["inputTarget"].read()
  File "<stdin>", line 1, in <module>
  File "D:\repos\cmdx\cmdx.py", line 4044, in read

  File "D:\repos\cmdx\cmdx.py", line 5076, in _plug_to_python
    #
  File "D:\repos\cmdx\cmdx.py", line 5094, in _plug_to_python
    )
  File "D:\repos\cmdx\cmdx.py", line 5094, in <genexpr>
    )
  File "D:\repos\cmdx\cmdx.py", line 5076, in _plug_to_python
    #
  File "D:\repos\cmdx\cmdx.py", line 5094, in _plug_to_python
    )
  File "D:\repos\cmdx\cmdx.py", line 5094, in <genexpr>
    )
  File "D:\repos\cmdx\cmdx.py", line 5119, in _plug_to_python
    # E.g. transform["worldMatrix"][0]
RuntimeError: (kFailure): Unexpected Internal Failure
```

Might be an edge case, but that was the reason why I added tat in.

I did some digging and it _could_ be because the attribute which it fails on (`blendshape1.inputTarget[0].inputTargetGroup[0].targetBindMatrix`) has its value being cached. It is just a theory at this point though, I haven't properly diagnosed this.
```python
# Isolated API fail:
>>> bs['inputTarget'][0][0][0][4]._mplug.asMObject()             
  File "<stdin>", line 1, in <module>
RuntimeError: (kFailure): Unexpected Internal Failure

>>> bs['inputTarget'][0][0][0][4]._mplug.isCaching
True
```

We could remove the try/except and deal with this as a separate issue though, which is probably a cleaner strategy.

cmdx.py
mottosso: > If this is not the case anymore, and we never expect str to fail, would it be ok to expect that it can return None? In the case of a Compound attribute for instance? That way I can add that check to include == or not.

Sure, let's try it.

> A third option, albeit more complex, could be to have it return a dictionary, with the child attribute names as keys for instance.

I mean, this would be supremely useful. But I'm uncertain how stable we can make this.. Maybe let's start small, and add this later on?

chelloiaco: > Sure, let's try it

Scratch that. We can't return something that isn't a string type from `__str__`. That raises a `TypeError`, silly me.

I think I was able to think of a decent solution within `__repr__`, so we're not messing with `__str__` in this PR. pushing a commit now which includes this and the [above](https://github.com/mottosso/cmdx/pull/78#discussion_r1500286153)

cmdx.py
```diff
@@ -2627,7 +2681,24 @@ def __str__(self):
         return str(self.read())
 
     def __repr__(self):
-        return str(self.read())
+        cls_name = '{}.{}'.format(__name__, self.__class__.__name__)
+        msg = '{}("{}", "{}")'.format(cls_name,
+                                      self.node().name(),
+                                      self.name())
+
+        if not all((self.isCompound, self.isArray)):
```
mottosso, line 2689: Do you mean `any()`? This means it can be an array, and it can be a compound, but it cannot be both at the same time

chelloiaco: `all` is correct. `any` would exclude any attribute that isn’t an array or compound.

The idea is that it can display an array _(in the case of `worldMatrix`, for example)_ and a compound _(in the case of `translate`)_. But, when dealing with a Compound that is an Array, that’s when it seems to start to become unclear what sort of data will be displayed, namely the example of `bs['input']` mentioned previously in this PR.

mottosso: > But, when dealing with a Compound that is an Array, that’s when it seems to start to become unclear what sort of data will be displayed, namely the example of bs['input'] mentioned previously in this PR

Ahaa, yes in that case. Good. Then this looks good I think! Well done, let's get this show on the road! 👏

